PR #11, opened 2019-12-06
Sync last viewed date from Plex to Trakt

fabiomsnunes:
Forgot the TV Shows, please replicate to line 220:
`seen_date = (episode.lastViewedAt if episode.lastViewedAt else datetime.now())
eps.instance.mark_as_seen(seen_date)`

### DIFF
--- a/main.py
+++ b/main.py
@@ -10,6 +10,7 @@ from dotenv import load_dotenv
 from os import getenv
 import logging
 from time import time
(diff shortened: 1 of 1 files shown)


gitthangbaby:
tested this and it works, it just adds +1 extra hour compared to time i see in Plex and system. thanks.

PR #13, opened 2019-12-28
improve configuration

twolaw:
Setup could be more user-friendly, here is what I did :

- .pytrakt.json is moved from home folder to project folder (with .env and config.json) so that all config files are together. Easier for maintenance or migration.
- if main.py runs for the first time, get_env_data.py is called.
- get_env_data.py writes PLEX_TOKEN and TRAKT_USERNAME in .env file, no need to manually copy/paste.

Also, why do you have `trakt.APPLICATION_ID = '65370'` line ?
It is only needed for pin auth method but you choose oauth method.
By the way, PIN authentication would be easier for such a basic script. Note that it is default pytrakt method.

### DIFF
--- a/get_env_data.py
+++ b/get_env_data.py
@@ -1,8 +1,11 @@
 from plexapi.myplex import MyPlexAccount
 import utils
+from os import path
 import trakt
 import trakt.core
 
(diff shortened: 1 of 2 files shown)


twolaw:
Great, you can now change setup instructions in Readme file.

Issue #20, opened 2020-03-27
Trakt token not being refreshed after 3 months


twolaw:
[Trakt API](https://trakt.docs.apiary.io/#reference/authentication-oauth/get-token/exchange-code-for-access_token) says `access_token` is valid for 3 months only. So after this time, the script can't connect to trakt api anymore and fail.

Solution : as the api say, the script must store and use the `refresh_token` to [get a new token](https://trakt.docs.apiary.io/#reference/authentication-oauth/get-token/exchange-refresh_token-for-access_token) when `access_token` expiration date comes.

Could you fix this please ?

cutzenfriend:
+1

twolaw:
My PR moogar0880/PyTrakt#121 has been merged into pytrakt 2.14 👍 
Therefore, updating trakt module requirement to version 2.14 fixes this issue.
@glensc I'm not familiar with pipfile.lock content and I don't want to break anything. Can you help me?

glensc:
@twolaw: https://github.com/Taxel/PlexTraktSync/pull/50

but I'm no heavy user of pipenv either, I just found it's useful in Python world as npm/yarn/composer lock in node/php worlds.

twolaw:
This issue is back, after this PR being merged :
- https://github.com/Taxel/PlexTraktSync/pull/314

Because of bad implementation of Trakt Auth Token processing in PyTrakt:
- https://github.com/moogar0880/PyTrakt/issues/150

Many decorators use, which I'm not understanding very well.

glensc:
Check that you don't have `.pytrakt.json` in the default location: `~/.pytrakt.json`. the freshness check may use that file instead of forced location of `.pytrakt.json`

glensc:
should be fixed with https://github.com/Taxel/PlexTraktSync/pull/715

please open a new bug report if it re-occurs or have more details.

PR #32, opened 2020-05-17
Switch to device auth

twolaw:
[Device authentication](https://trakt.docs.apiary.io/#reference/authentication-devices) is easier to setup for users with command line scripts.
Users don't have to copy/paste a long url but only a code.
Moreover, the pytrakt function of device_auth stores the `oauth_refresh` token which can be used to automatically refresh the `oauth_token` every 3 months.

### DIFF
--- a/get_env_data.py
+++ b/get_env_data.py
@@ -21,10 +21,9 @@ else:
     with open(env_file, "w") as txt:
         txt.write("PLEX_TOKEN=-\n")
 
-trakt.APPLICATION_ID = '65370'
-trakt.core.AUTH_METHOD=trakt.core.OAUTH_AUTH
+trakt.core.AUTH_METHOD=trakt.core.DEVICE_AUTH
 trakt_user = input("Please input your Trakt username: ")
-trakt.init(trakt_user, store=True)
+trakt.init(store=True)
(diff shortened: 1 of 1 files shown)


glensc:
forgotten cross reference: https://github.com/Taxel/PlexTraktSync/issues/20

get_env_data.py
```diff
@@ -21,10 +21,9 @@
     with open(env_file, "w") as txt:
         txt.write("PLEX_TOKEN=-\n")
 
-trakt.APPLICATION_ID = '65370'
-trakt.core.AUTH_METHOD=trakt.core.OAUTH_AUTH
+trakt.core.AUTH_METHOD=trakt.core.DEVICE_AUTH
 trakt_user = input("Please input your Trakt username: ")
-trakt.init(trakt_user, store=True)
+trakt.init(store=True)
```
glensc: `trakt_user` is no longer used for trakt auth. do we need to keep asking username from the user?

twolaw: You're wrong, it is still used. Read main.py

glensc: I'm not wrong, it's no longer used for trakt auth in that file. that's why asked is it not used elsewhere as well.

PR #38, opened 2020-06-14
Add more specific install instructions (Alpine for now)

glensc:


glensc:
@twolaw wdyt?

twolaw:
I think it's too specific and not really needed.
The average user (on Windows, Linux or Mac) may be disturbed.

glensc:
for macOS (brew) nothing is needed (i don't recall needing to install anything extra), other Linux distros could be added as well, and I'm assuming nothing is needed on windows either otherwise it would be already documented?

Issue #51, opened 2020-07-14
Give user indication what to do on error


glensc:
I've updated to recent master (39c89b9) and now I get error:

```
root@acai:~/scm/mediaserver/plex/Plugins/PlexTraktSync # ./plex_trakt_sync.sh
Loading .env environment variables…
Starting sync Plex None and Trakt *censored*
Trakt authentication error: Unauthorized - OAuth must be provided
```

I did ran `pip3 install -r requirements.txt` to grab latest packages, the error is still there

I really would not wish to dig through merged pull requests to see what changed and how to log in again. I think the error should give me instructions on what to do next.

twolaw:
Remove all files and start over the setup process.

glensc:
I mean the error is clear: Unauthorized, so tell user how to authenticate.

twolaw:
No, it's not normal use case.
This is an already fixed bug, you used an old trakt oauth token which doesn't come with a refresh token.
Wipe everything as I said and restart over.

glensc:
What do you mean everything? I don't remember how I started it. And creating new trakt app again does not seem neccessary.

Removing `.pytrakt.json` should be sufficient?

glensc:
Seems still need to create new app??!

```
root@acai:~/scm/mediaserver/plex/Plugins/PlexTraktSync # mv .pytrakt.json .pytrakt.json.old
root@acai:~/scm/mediaserver/plex/Plugins/PlexTraktSync # ./plex_trakt_sync.sh
Loading .env environment variables…
Starting sync Plex None and Trakt **censored**
Traceback (most recent call last):
  File "/home/glen/scm/mediaserver/plex/Plugins/PlexTraktSync/main.py", line 341, in <module>
    main()
  File "/home/glen/scm/mediaserver/plex/Plugins/PlexTraktSync/main.py", line 270, in main
    trakt_user = trakt.users.User(getenv('TRAKT_USERNAME'))
  File "/home/glen/.local/share/virtualenvs/PlexTraktSync-sKH3BWbK/lib/python3.7/site-packages/trakt/users.py", line 222, in __init__
    self._get()
  File "/home/glen/.local/share/virtualenvs/PlexTraktSync-sKH3BWbK/lib/python3.7/site-packages/trakt/core.py", line 472, in inner
    json_data = self._handle_request('get', url)
[...]

PR #52, opened 2020-07-25
Update get_env_data to re-use exiting information

glensc:
Resolves https://github.com/Taxel/PlexTraktSync/issues/51

### DIFF
--- a/get_env_data.py
+++ b/get_env_data.py
@@ -1,18 +1,29 @@
 from plexapi.myplex import MyPlexAccount
+from dotenv import load_dotenv
 import utils
-from os import path
+from os import getenv, path
 import trakt
 import trakt.core
 
 trakt.core.CONFIG_PATH = path.join(path.dirname(path.abspath(__file__)), ".pytrakt.json")
 env_file = path.join(path.dirname(path.abspath(__file__)), ".env")
 
+load_dotenv()
+PLEX_BASEURL = getenv("PLEX_BASEURL")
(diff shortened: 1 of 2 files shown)


twolaw:
Can you describe in few words what this PR is supposed to do ?

I don't understand why you load trakt CLIENT_ID and CLIENT_SECRET. And you never use it later.
```
PLEX_USERNAME = username
PLEX_TOKEN = token
```
2 variables for the username and 2 variables for the token. I think it's not good. And if username is not set it will raise an error.

glensc:
> Can you describe in few words what this PR is supposed to do?

as the title says, and the issue links to, when invoking the script, re-use existing values instead of having the user to remember the details (or lookup from the wallet)

> 2 variables for the username and 2 variables for the token. I think it's not good. And if username is not set it will raise an error.

it's just consistent to ensure whatever the code path ends up, the value gets assigned, as the code above is way complex (should be moved to function for isolated context). 

> I don't understand why you load trakt CLIENT_ID and CLIENT_SECRET. And you never use it later.

I've rebased this branch so many times, that the original view is kind of lost. I've left them for future use.

twolaw:
The new user will see (None) which is weird.
It still raises an error because of PLEX_USERNAME = username as I told you.
> re-use existing values instead of having the user to remember the details

As I said updating credentials is very exceptionnal.
Personally, I disapprove this PR because I think it's not needed, not clean and not working good.

PR #53, opened 2020-07-29
Use input_hidden to ask for password

glensc:
Extracted out of https://github.com/Taxel/PlexTraktSync/pull/52

### DIFF
--- a/get_env_data.py
+++ b/get_env_data.py
@@ -11,7 +11,7 @@ env_file = path.join(path.dirname(path.abspath(__file__)), ".env")
 plex_needed = utils.input_yesno("-- Plex --\nAre you logged into this server with a Plex account?")
 if plex_needed:
     username = input("Please enter your Plex username: ")
-    password = input("Please enter your Plex password: ")
+    password = utils.input_hidden("Please enter your Plex password: ")
     servername = input("Now enter the server name: ")
     account = MyPlexAccount(username, password)
     plex = account.resource(servername).connect()  # returns a PlexServer instance
--- a/utils.py
+++ b/utils.py
@@ -1,5 +1,12 @@
 # helper functions
 
+import getpass
+
+def input_hidden(msg):
+    reply = getpass.getpass(prompt = msg)
+
+    return reply
+
 def input_yesno(msg):
     answer = str(input(msg + ' (y/n): ')).lower()
     if answer in ['yes', 'y']:


twolaw:
Did you check if password is asked again after a wrong password input ?
If the script crashes when the user misspell his password (because he can't see what's input), it will be frustrating for him.

glensc:
How could this function know if password is incorrect?

This would require refactoring in calling part to ask for input in a loop, and even then the code can't know if password was incorrect or username, so it should ask everything again.

The only thing usable in this function is to print prompt again if user entered empty password.

glensc:
Also, who would ever want to see their password? I'm rather surprised that this tool shows password when I type.

twolaw:
> the code can't know if password was incorrect or username, so it should ask everything again.

I agree. Asking both username and password again is a good solution.

twolaw:
Maybe a loop like I did for managed users, to ask credentials again and again until Plex account is found.

glensc:
@twolaw can you link to your change what you had in mind? I will look onto this later sometime.

twolaw:
Something like this :

```python
while True:
    try:
        username = input("Enter your Plex username: ")
        password = utils.input_hidden("Enter your Plex password: ")
        account = MyPlexAccount(username, password)
        break
    except plexapi.exceptions.BadRequest as e:
        print("Error: {}".format(str(e))
        continue
while True:
    try:
        servername = input("Enter the server name: ")
        plex = account.resource(servername).connect()  # returns a PlexServer instance
        break
    except plexapi.exceptions.NotFound as e:
        print("Error: {}".format(str(e))
        continue
```

glensc:
I expected you to link to a commit, but o-kay

get_env_data.py
```diff
@@ -11,7 +11,7 @@
 plex_needed = utils.input_yesno("-- Plex --\nAre you logged into this server with a Plex account?")
 if plex_needed:
     username = input("Please enter your Plex username: ")
-    password = input("Please enter your Plex password: ")
+    password = utils.input_hidden("Please enter your Plex password: ")
```
twolaw, line 14: As said before, it should catch wrong credentials and ask again.

twolaw (changes requested):

get_env_data.py
glensc: @twolaw I have not made any changes since the last review.

I only rebased against the current master, maybe you go notification on the rebase push.

glensc:
This PR is no longer relevant.

Added new command to handle plex login:
- https://github.com/Taxel/PlexTraktSync/pull/230

PR #58, opened 2020-08-20
keep trakt rank when sync playlists to plex

### LINKED ISSUE #54: Sync not considering rank
Hi, 

This script was exactly what I was looking for. Nevertheless it does not consider the rank position on the list e.g. when I reorder my watchlist and sync again, the playlist in Plex does not update accordingly. I'm I doing something wrong or can this be added?

twolaw:
This makes the script respect the Trakt playlists order (items rank is preserved) when syncing them to Plex.

fixes #54

### DIFF
--- a/main.py
+++ b/main.py
@@ -284,7 +284,7 @@ def main():
         #logging.debug("Movie collection from trakt:", trakt_movie_collection)
         trakt_watched_shows = pytrakt_extensions.allwatched()
         if CONFIG['sync']['watchlist']:
-            listutil.addList(None, "Trakt Watchlist", list_set=set(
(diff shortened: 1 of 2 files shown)


Taxel:
List should be slower for lookup when using many/huge trakt lists (with a list it's trivial to solve, though). Since this comparison happens at every movie this might be significant.
The code looks fine, if it works for you and does not increase the runtime by a lot, go for it and merge it.

Taxel (approved):
Looks fine. As long as it works, go for it 👍

glensc:
Extra discussion about how sort works:
- https://github.com/Taxel/PlexTraktSync/pull/1184#discussion_r1018449468

PR #60, opened 2020-08-23
Update run script to rotate logs

glensc:
This modifies `plex_trakt_sync.sh` script to keep sepatate log for each run:

- "update-$date.log" - for success
- "error-$date.log" - for errors

### DIFF
--- a/plex_trakt_sync.sh
+++ b/plex_trakt_sync.sh
@@ -1,3 +1,13 @@
 #!/bin/sh
 PATH=/usr/local/bin:/usr/local/sbin:~/bin:/usr/bin:/bin:/usr/sbin:/sbin
-python3 ./main.py
+set -eu
+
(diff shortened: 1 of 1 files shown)


twolaw:
I think it's better if the last update log always have the same name.
So that we can easily read it with a simple `tail last_update.log` command and so it can be fetch by other logging tools.
Only the older ones would be renamed and archived.

As far as I'm concerned, I just changed (in main.py) the `write` argument of log FileHandler to an `append` argument. I like having all in the same file.

leipert:
I would generally recommend logrotate (https://linux.die.net/man/8/logrotate) to rotate logs 😄

glensc:
@leipert then one needs to synchronize logrotate invocation with sync script, to be able to have a separate log for each run.

PR #70, opened 2020-09-20
Allow custom envfile path to be used

docmeth02:
This will allow to supply a custom envfile to the script. Im using PlexTraktSync to keep three servers in sync and it works perfectly. This change makes it very easy to reuse the script to sync multiple servers. There is no verification or checks in place the custom envfile really exists. Its not needed for my purposes. If you want this added feel free to ping me and i implement it :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,12 @@ by adding the line:
 
 to the cronjobs after typing `crontab -e` in the terminal.
 
+You can supply a custom env file by using the CUSTOM_ENV envoirment variable. 
+
+```
+CUSTOM_ENV=.env.my.plex.host ./plex_trakt_sync
+```
+
 ## Sync settings
 
 To disable parts of the functionality of this software, look no further than
--- a/main.py
+++ b/main.py
@@ -254,7 +254,15 @@ def process_show_section(s, watched_set):
 def main():
 
     start_time = time()
-    load_dotenv()
+    # try to load the custom envfile set in CUSTOM_ENV envoirment variable
+    envFile = getenv('CUSTOM_ENV')
+    if envFile == None:
+        envFile = path.join(path.dirname(path.abspath(__file__)), '.env')
+    else:
+        if not path.isabs(envFile):
+           envFile = path.join(path.dirname(path.abspath(__file__)), envFile) 
+        print("Using custom env file: %s" % envFile)  
+    load_dotenv(dotenv_path=envFile)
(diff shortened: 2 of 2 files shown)


glensc:
@docmeth02 can you make this backward compatible? instead of modifying execution syntax, use some environment variable to load environment variables?

or why not just solve this in the shell itself?

```
(. ./env1 && python3 sync.py)
(. ./env2 && python3 sync.py)
(. ./env3 && python3 sync.py)
```

docmeth02:
@glensc good idea! I updated the PR to use a environment variable.

twolaw:
I suggest to path.joint with `path.dirname(path.abspath(__file__))` instead of `'.'` which is more robust (if user run python script with absolute path).

docmeth02:
good catch @twolaw
i made sure to to the same with the custom env file if a non absolute path is supplied!

glensc:
@docmeth02 the argv import is now superfluous. also there's change in .sh sync script that doesn't change anything. perhaps you want to git rebase, squash your commits and force push over this branch.

docmeth02:
done @glensc

JpF4E:
If we isolated both the .env and .pytrakt.json files (i.e allowed a custom config dir), we could support multiple users (different plex and trakt users), am I right?

twolaw:
> 
> 
> If we isolated both the .env and .pytrakt.json files (i.e allowed a custom config dir), we could support multiple users (different plex and trakt users), am I right?

You would also have to isolate config.json.
@VladButacu already worked on this here : https://github.com/VladButacu/PlexTraktSync
He isolated config files in a data folder.

JpF4E:
@twolaw shouldn't this merged to this repo as a PR? It seems like a way of having multiple users...

twolaw:
The incoming file reorganisation #89 with a dedicated config folder will offer a better way to do it I think.
Allowing not only multiple plex users but also multiple trakt users.

glensc:
@docmeth02 can we revisit this, why do you actually need this?

perhaps similarly to https://github.com/Taxel/PlexTraktSync/pull/191 add `--server` commandline option?

glensc:
This targets the old unsupported 0.5.x branch and still has conflicts.

If this is still needed, submit a new pull request. but first consider maybe it should be implemented differently:
- https://github.com/Taxel/PlexTraktSync/pull/70#issuecomment-814845033

PR #82, opened 2020-11-26
Support for Plex Movie Scanner

wampa1:
I think I figured out how to do a pull request. Forked it, then made the change then submit the request I guess.

This uses name and year matching to attempt to mitigate the plex movie scanner guid misalignment with Trakt's available lookups.

Hope this helps.  It definitely helped me (and worked like a champ).

Eric

### DIFF
--- a/main.py
+++ b/main.py
@@ -54,6 +54,13 @@ def process_movie_section(s, watched_set, ratings_dict, listutil, collection):
             x = guid.split('//')[1]
             x = x.split('?')[0]
             provider = CONFIG['xbmc-providers']['movies']
+        elif 'plex:' in guid:
+            # We can't use the guid for this as it's not going to match Trakt.
+            x = movie.title
+            provider = 'plex'
+            # We'll do a name based query against Trakt for the Plex Movie Scanner
+            # It's a lot looser, but we'll do our best to match it up.
+            # It's more likely we'd get misses than false positives.
         else:
             logging.error('Movie [{} ({})]: Unrecognized GUID {}'.format(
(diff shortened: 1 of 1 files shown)


glensc:
@wampa1 in the future, create a branch when submitting PR. you may want your `master` to live it's own life, but it's directly linked to specific PR. any commits you add to branch linked to the PR, will add more commits to the PR. in other words, never create pull request from `master` branch.

main.py
```diff
@@ -54,14 +54,31 @@ def process_movie_section(s, watched_set, ratings_dict, listutil, collection):
...
         else:
             logging.error('Movie [{} ({})]: Unrecognized GUID {}'.format(
                 movie.title, movie.year, movie.guid))
             continue
             raise NotImplementedError()
         # search and sync movie
         try:
-            search = trakt.sync.search_by_id(x, id_type=provider)
+            search = []
+            if ( provider == 'plex' ):
```
glensc, line 72:
```suggestion
            if provider == 'plex':

```

superfluous parens can be dropped

wampa1: If it's cool with you, I'd like to follow both of your suggestions.

I'll remove the parens (I am 99.99% a Java guy), kill this PR, and start a new one that's from a non-master branch.

I hope killing the PR and starting a new one doesn't cause you any additional concern.

Thanks again!

glensc: you can force push to the same branch you created the pr, and PR will update. no need to create new PR.

you can also apply the suggestion from web. there's UI for that. but you want then probably git squash the commit locally (and the force push).

PR #84, opened 2020-12-02
Add rate limit detection and looping

jbalcorn:
Code will detect a rate limit from Trackt and will sleep and loop until the attempted update is complete.

### DIFF
--- a/main.py
+++ b/main.py
@@ -10,7 +10,7 @@ import trakt.users
 import trakt.core
 from dotenv import load_dotenv
 import logging
-from time import time
+from time import time, sleep
 import datetime
 from json.decoder import JSONDecodeError
 
@@ -131,9 +131,13 @@ def process_movie_section(s, watched_set, ratings_dict, listutil, collection):
         except trakt.errors.NotFoundException:
             logging.error(
                 "Movie [{} ({})]: GUID {} not found on trakt".format(movie.title, movie.year, guid))
-        except:
+        except trakt.errors.RateLimitException:
+            logging.warning(
+                "Movie [{} ({})]: Rate Limited. Sleeping 1 sec from trakt (GUID: {})".format(movie.title, movie.year, guid))
+            sleep(1)
+        except Exception as e:
             logging.error(
-                "Movie [{} ({})]: bad response from trakt (GUID: {})".format(movie.title, movie.year, guid))
+                "Movie [{} ({})]: Response '{}' from trakt (GUID: {})".format(movie.title, movie.year, e.message, guid))
 
 
 def process_show_section(s, watched_set, listutil):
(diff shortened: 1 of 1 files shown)


glensc (approved):

glensc:
While I'd rather added a helper to make code blocks retriable, but the current code base is so horrible already, not worth the effort.

I'm not a daily Python coder, but I think python had `with` keyword that you could help in this area.

In fact requests cache is used in code in similar manner:

```python
    with requests_cache.disabled():
        sections = plex.library.sections()
```

glensc:
Probably related to https://github.com/trakt/api-help/issues/220

glensc:
And to do the rate-limiting correctly, you should read the `X-Ratelimit` header value to obtain how long to sleep:
- https://trakt.docs.apiary.io/#introduction/rate-limiting

This is to someone if they wish to improve this in the future.

main.py
```diff
@@ -131,9 +159,13 @@ def process_movie_section(s, watched_set, ratings_dict, listutil, collection):
...
             logging.error(
                 "Movie [{} ({})]: GUID {} not found on trakt".format(movie.title, movie.year, guid))
-        except:
+        except trakt.errors.RateLimitException:
+            logging.warning(
+                "Movie [{} ({})]: Rate Limited. Sleeping 1 sec from trakt (GUID: {})".format(movie.title, movie.year, guid))
+            sleep(1)
+        except Exception as e:
             logging.error(
-                "Movie [{} ({})]: bad response from trakt (GUID: {})".format(movie.title, movie.year, guid))
```
twolaw, line 136: This "bad response" string is because sometimes trakt responses are not correct but don't raise a trakt error.
It is raised as a json error by json decoder :
`json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

So the e.message is not a response from trakt.

jasantunes:
I'm using this to work around the rate limiting. Thanks!

main.py
twolaw: I'd like to fix this but my push fails with _permission denied_ from jbalcorn repository.

twolaw:
> And to do the rate-limiting correctly, you should read the `X-Ratelimit` header value to obtain how long to sleep:
> 
>     * https://trakt.docs.apiary.io/#introduction/rate-limiting
> 
> This is to someone if they wish to improve this in the future.

The header value `Retry-After` should be used for this.

And I think there is a problem of infinite loop if for any reason Trakt API keep sending the RateLimitException response.

doctarrwho:
Are there any plans of merging this PR at some point?

JpF4E:
A develop branch would be nice 🙂

main.py
glensc: @twolaw I've created new PR based on this PR:
- https://github.com/Taxel/PlexTraktSync/pull/119

you can push and finish up there.

glensc:
rebased version at https://github.com/Taxel/PlexTraktSync/pull/119

Issue #88, opened 2020-12-08
Docker support


Taxel:
As you might have seen, I have pushed the _work in progress_ docker branch. I am currently testing the built image in my home setup (with this image as well as Plex running in Kubernetes administered via Rancher) but I have made some changes to how the script works, files have been reorganized etc. so I wanted to discuss them here first before even submitting a pull request:

- I rebased the branch on my changes in #87, if we wait for the official python-plexapi package to get updated this will have to be redone
- I changed the file structure: With configuration, cache, logs, and source code in one directory the project felt unorganized, so I should have probably done this before even making the project open source.
- The docker image is based on @twolaw 's Dockerfile. The main differences are, that I am using a multistep build (unnecessarily I might add, this does not really make the resulting image much smaller, but it was fun to play around with) and that the script does not run once by default but on a cronjob, so the container can be configured and then left running forever. One can, however, pass the environment variable "RUN_ONCE" with a non-empty value and the script will only be run once, after which the container shuts down. Changing the cron schedule is also done via an environment var: CRON_SCHEDULE
- Running this in Docker also means that via volume mounts and multiple containers, one can use different configs for multiple users and/or Plex servers, solving #70 indirectly (only in Docker, of course. Not when running this with Python. Since I extracted all the filepaths to config.py this should be trivial to implement, though)

What might be the most controversial change, is that I am rerouting the log outputs to stdout if the terminal is non-interactive. I use it to detect if the script is running in a container since Docker containers should output to stdout/stderr and let Docker take care of the logging. Do you agree the script should no longer write to `lastRun.log` if it is running in a container and is the `sys.isatty(0)` check okay or should I add an argument like `python3 main.py --log-to-stdout` so this is easier to control?

Since the file structure changes alone constitute some manual work when upgrading from an older version of PlexTraktSync, I also propose we create a release 0.9.0 before merging the pull request for this once the feature is complete and then release 1.0.0 along with the docker image (I'd like to try Github Packages as the image registry and create a Github action that automatically builds, tags and pushes each new release in the future)

@glensc @twolaw what do you think about these changes? If you want to test them out locally it should suffice to `docker-compose up --build` after pulling the branch. Are there other things I should add to the image prior to release (for example atm the container always runs as root. I didn't deem it necessary to make it possible to change this since the container just mounts the config directory and is not accessible from the outside because it does not listen for incoming connections)

Greez
Taxel

glensc:
[...]

PR #87, opened 2020-12-08
Update to use guids from python-plexapi to fix Plexmovieagent breakage

Taxel:
Hey guys,

after finally getting to installing this script in my home, I was immediately annoyed by the fact that the new plex movie agent does not work so far.
So I went ahead and included the tar.gz of the [pull request in python-plexapi that fixes the issue](https://github.com/pkkid/python-plexapi/pull/562) and some minor modifications that use the new `movie.guids` property if the new Plex movie agent is used.

It works fine on my machine, if someone who still has a library with the old movie agent can confirm I didn't break anything there, we can merge this.

This fixes #56 and makes #83 unnecessary because it turns out the newer plexapi actually gives back the imdb and/or tvdb links in the new property, no more "search by name and year and hope it's the right match" necessary.

Once the pull request on the `python-plexapi` package is merged, we should be able to go to the latest version there without problems.

Also, I just deleted and recreated the Pipfile and Pipfile.lock, which apparently pinned Python 3.8 as the version. Since I have no experience with this: Is there a better way to do this?

## Installation
Simply reinstalling your dependencies, either with `pip3 install -r requirements.txt` or `pipenv install` should be all that's needed.

Greez
Taxel

----

Replaces https://github.com/Taxel/PlexTraktSync/pull/86 which was unintentionally closed.

Fixes https://github.com/Taxel/PlexTraktSync/pull/83 (replaces)

----

@glensc I followed all your suggestions, splitting up the commit, using Python 3 instead of Python 3.8 in Pipenv, and accidentally pushing to master in the process. The master can now no longer be pushed to (even by me) to avoid this in the future.

This apparently closed the other PR, so here's a new one.

PR #89, opened 2020-12-08
WIP: Docker support

Taxel:
As per note of @glensc in #88  this is now a pull request.  

The following changes have been made:

- I rebased the branch on my changes in #87, if we wait for the official python-plexapi package to get updated this will have to be redone
- I changed the file structure: With configuration, cache, logs, and source code in one directory the project felt unorganized, so I should have probably done this before even making the project open source.
- The docker image is based on @twolaw 's Dockerfile. The main differences are, that I am using a multistep build (unnecessarily I might add, this does not really make the resulting image much smaller, but it was fun to play around with) and that the script does not run once by default but on a cronjob, so the container can be configured and then left running forever. One can, however, pass the environment variable "RUN_ONCE" with a non-empty value and the script will only be run once, after which the container shuts down. Changing the cron schedule is also done via an environment var: CRON_SCHEDULE
- Running this in Docker also means that via volume mounts and multiple containers, one can use different configs for multiple users and/or Plex servers, solving #70 indirectly (only in Docker, of course. Not when running this with Python. Since I extracted all the filepaths to config.py this should be trivial to implement, though)

What might be the most controversial change, is that I am rerouting the log outputs to stdout if the terminal is non-interactive. I use it to detect if the script is running in a container since Docker containers should output to stdout/stderr and let Docker take care of the logging. Do you agree the script should no longer write to `lastRun.log` if it is running in a container and is the `sys.isatty(0)` check okay or should I add an argument like `python3 main.py --log-to-stdout` so this is easier to control?

Since the file structure changes alone constitute some manual work when upgrading from an older version of PlexTraktSync, I also propose we create a release 0.9.0 before merging the pull request for this once the feature is complete and then release 1.0.0 along with the docker image (I'd like to try Github Packages as the image registry and create a Github action that automatically builds, tags and pushes each new release in the future)


entry.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+# copy /var/config.json to /var/config/config.json if it does not exist
+if [ ! -f /var/config/config.json ]; then
+    echo "config.json does not exist. copying it."
+    cp /var/config.json /var/config/config.json
+fi
+
+if [[ ! -z "${RUN_ONCE}" ]]; then
```
glensc, line 9: `[[` is bashism, but `/bin/sh` is the interpreter.

plex_trakt_sync.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 PATH=/usr/local/bin:/usr/local/sbin:~/bin:/usr/bin:/bin:/usr/sbin:/sbin
-python3 ./main.py
\ No newline at end of file
+(cd ./src && python3 ./main.py)
```
glensc, line 3: you don't need a subshell, there's nothing else after `python` command, so no need to subshell `cd`'ing

in fact. you should use `exec` to release memory from shell process:

```suggestion
cd ./src && exec python3 ./main.py
```

and perhaps just use `set -e` instead of `&&`

```suggestion
see -e
cd ./src
exec python3 ./main.py
```

entry.sh
```diff
@@ -0,0 +1,26 @@
...
+
+if [[ -z "${CRON_SCHEDULE}" ]]; then
+  echo "Running the script every 2 hours (CRON_SCHEDULE='0 */2 * * *'). Set environment var 'CRON_SCHEDULE' to a valid cron expression to configure this"
+  echo "If you want to run the script only once, specify the environment var 'RUN_ONCE=1'"
+  SCHEDULE="0 */2 * * *"
+else
+  SCHEDULE="${CRON_SCHEDULE}"
+fi
+# install cronjob
+(crontab -l ; echo "$SCHEDULE python3 /var/src/main.py") | sort - | uniq - | crontab -
```
glensc, line 24: you did `cd` in `plex_track_sync.sh`, but not here. is the `cd` needed or not?

glensc: `| sort - | uniq - ` can be replaced with `| sort -u`

but check it first, you may have crippled `sort` in your image, i.e one from (older) `busybox`

entry.sh
```diff
@@ -0,0 +1,26 @@
...
+# run crond so the container does not terminate
+crond -l 2 -f
```
glensc, line 26: if this `crond` is last command in the script, prepend with `exec`, so that:
1. memory is released from shell
2. signals get routed to main process, not shell (which typically ignores them)

twolaw:
I ran `docker-compose up -d` and the image built fine (on arm: 75MB, mine was 58MB).
But the container is created without /var/config/ folder so the entry.sh can"t go further than :
```
config.json does not exist. copying it.
cp: can't create '/var/config/config.json': No such file or directory
Running the script every 2 hours (CRON_SCHEDULE='0 */2 * * *'). Set environment var 'CRON_SCHEDULE' to a valid cron expression to configure this
If you want to run the script only once, specify the environment var 'RUN_ONCE=1'
Traceback (most recent call last):
  File "/var/src/main.py", line 3, in <module>
    from config import CONFIG, CONFIG_DIR, LOG_DIR, CACHE_DIR
  File "/var/src/config.py", line 11, in <module>
    with open(config_file, 'r') as fp:
FileNotFoundError: [Errno 2] No such file or directory: '/var/src/../config/config.json'
```
Is the volume map for config/ missing in docker-compose.yml ?
Did I do something wrong ?

SJ50:
@twolaw you need to mount “src” and “config” folder. 

./src:/var/src
./config:/var/config

PR #114, opened 2021-02-28
Use pip3 for pipenv as well

glensc:


twolaw (approved):

twolaw:
Hi glensc, I approve only PR I am sure to fully understand the benefits.
Taxel remains the leader here.
Just like you, I don't understand why many things are started and not finished (files structure reorganised, trakt requests limit, new plex agent, docker,...)

Issue #95, opened 2020-12-21
TV library sync reporting "bad response from trakt"


doctarrwho:
Script runs fine, but TV library sync continues to report a lot of failures. Rerunning the script collects/syncs some more shows/episodes but it fails again. It almost feels like the calls are getting throttled.

2020-12-21 02:27:03,167 INFO:Show [Bless This Mess (2019)]: Marked as watched on trakt: episode S02E07
2020-12-21 02:27:03,325 INFO:Show [Bless This Mess (2019)]: Collected episode S02E08
2020-12-21 02:27:03,490 INFO:Show [Bless This Mess (2019)]: Marked as watched on trakt: episode S02E08
2020-12-21 02:27:03,575 ERROR:Show [Bless This Mess (2019)]: bad response from trakt (GUID com.plexapp.agents.thetvdb://357882?lang=en)
2020-12-21 02:27:06,321 ERROR:Show [The Boys (2019)]: bad response from trakt (GUID com.plexapp.agents.thetvdb://355567?lang=en)
2020-12-21 02:27:07,062 INFO:Show [Brave New World (2020) (2020)]: Finished sync
2020-12-21 02:27:07,985 INFO:Show [Breaking Bad (2008)]: Collected episode S02E05
2020-12-21 02:27:08,172 INFO:Show [Breaking Bad (2008)]: Collected episode S02E06
2020-12-21 02:27:08,306 ERROR:Show [Breaking Bad (2008)]: bad response from trakt (GUID com.plexapp.agents.thetvdb://81189?lang=en)
2020-12-21 02:27:09,290 INFO:Show [Breathe (2018)]: Collected episode S01E08
2020-12-21 02:27:09,422 ERROR:Show [Breathe (2018)]: bad response from trakt (GUID com.plexapp.agents.thetvdb://341337?lang=en)
2020-12-21 02:27:10,336 ERROR:Show [Broad City (2014)]: bad response from trakt (GUID com.plexapp.agents.thetvdb://275557?lang=en)
2020-12-21 02:27:11,082 INFO:Show [The Butcher (2019) (None)]: Finished sync
2020-12-21 02:27:11,787 ERROR:Show [Castle Rock (2018)]: bad response from trakt (GUID com.plexapp.agents.thetvdb://331600?lang=en)

I did setup the script yesterday, and confirmed that it has the change from #71.

[...]

PR #119, opened 2021-03-05
Add rate limit detection and looping

glensc:
Carries @jbalcorn's https://github.com/Taxel/PlexTraktSync/pull/84:
> Code will detect a rate limit from Trackt and will sleep and loop until the attempted update is complete.

Fixes https://github.com/Taxel/PlexTraktSync/issues/95

Rebased to current `master`


glensc:
@twolaw did you want to make changes on top of these?
- https://github.com/Taxel/PlexTraktSync/pull/84#discussion_r553220390

you can make the changes here now:
- https://github.com/Taxel/PlexTraktSync/pull/84#discussion_r588757783

twolaw:
> 
> @twolaw did you want to make changes on top of these?


Yes, I'll find time to polish this PR.

twolaw:
@glensc I don't know how to put #129 code inside this PR, I didn't want to create a separated PR.
If you can help me here, thanks.

glensc:
@twolaw can we merge this now?

ps: I'll not review this code, it's too big noise ratio to me, but I can give approval as i see it works for you.

glensc (commented):
duh: i can not self-review.

twolaw (approved):

twolaw:
@glensc tag 0.3.0 ?

jbalcorn:
> Important : the Retry-After value can be read from response only when [this PR](https://github.com/moogar0880/PyTrakt/pull/138) of PyTrakt will be merged. So we should **wait before merging this**.
> 
This explains why I never saw the Retry-After in my testing.

glensc:
@twolaw what is the point of `respect_trakt_rate` / `trakt_post_wait`?

Isn't `Retry-After` the number of seconds from the current time to wait before retry?

If the goal is to set a minimum delay for the wait to be 1.2s, which can be accomplished just `Math.max(header[Retry-After], 1.2)`? am I missing something?

Also, I don't understand the comment about `POST` requests. these do not have `Retry-After` header?

glensc:
@twolaw ah, I see now:

- https://trakt.docs.apiary.io/#introduction/rate-limiting

The limit for POST is super low, 1 call per second, so would immediately hit that limit:

Name | Verb | Methods | Limit
-- | -- | -- | --
AUTHED_API_POST_LIMIT | POST, PUT, DELETE | all | 1 call per second
AUTHED_API_GET_LIMIT | GET | all | 1000 calls every 5 minutes
UNAUTHED_API_GET_LIMIT | GET | all | 1000 calls every 5 minutes

twolaw:
Yes, the point is to avoid rate-limit errors as explained in my previous comment :

> respect POST requests interval to avoid rate-limit errors (instead of going full speed and get many 429 errors)

Maybe it can be improved but I found that `LIMIT + 0.2 sec` is a very good value to get speed and rate-limit respected.

glensc:
@twolaw it's good, I just did not understand it at first.

added the same to the decorator:
- https://github.com/Taxel/PlexTraktSync/pull/177/files

twolaw:
Idea of improvement (not tested) :
We could remove this 1.2 sec hardcoded value (I admit it's not perfect) and go full speed, then at first rate-limit-error raised we could read the Retry-After and set our delay at this value + 0.2
So if one day Trakt change the 1sec value to a lower or higher value, our script would adapt automatically.

main.py
```diff
@@ -72,14 +73,27 @@ def process_movie_section(s, watched_set, ratings_dict, listutil, collection):
...
+                            m.add_to_library()
+                            logging.info('Movie [{} ({})]: Added to trakt collection'.format(
+                                movie.title, movie.year))
+                            break
+                        except trakt.errors.RateLimitException as e:
+                            delay = int(e.response.headers.get("Retry-After", 1))
+                            logging.warning(
+                                "Movie [{} ({})]: Rate Limited on adding to collection. Sleeping {} sec from trakt (GUID: {})".format(movie.title, movie.year, delay, guid))
+                            sleep(delay)
+                            retry += retry
```
glensc, line 93: what does `retry += retry` supposed to do? seems to be `retry += 0` here as retry is `0` initially, and I don't see where it gets a different value.

twolaw: It's a mistake, it should be `retry += 1`
When it reaches 5 it exits the loop.

glensc: OK. clear!

twolaw: fixed in PR #187

PR #116, opened 2021-02-28
Move code to plex_trakt_sync module

glensc:
This adds some structure to the project, following documentation:
- https://docs.python-guide.org/writing/structure/

# details
- Moved code to `plex_trakt_sync`
- Add `path` module dealing with locations of the files
- Kept entrypoints (as wrappers):
  -  `clear_trakt_collections.py`
  -  `get_env_data.py`
  -  `main.py`

This is backward compatible change (should not change anything to users who update with git pull). Tested with local sync which completed OK.

PR #120, opened 2021-03-06
Update config module to lazy initialize and include environment variables

glensc:
This consolidates access to "config" type items to one `CONFIG` symbol. Makes it possible to add new config keys without much duplication. Also, the initialization is delayed so "first install" dialogs are invoked when config is accessed, not when module is loaded. In the future could add data key specific validation or defaults. And overloading config keys from environment.

NOTE:
- [x] Built on top of https://github.com/Taxel/PlexTraktSync/pull/116, so it needs to be merged first.

### DIFF
--- a/plex_trakt_sync/config.py
+++ b/plex_trakt_sync/config.py
@@ -1,5 +1,20 @@
 import json
 from plex_trakt_sync.path import config_file
 
-with open(config_file, 'r') as fp:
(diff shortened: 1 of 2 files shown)


glensc:
@twolaw ready for review

twolaw (approved):

PR #129, opened 2021-03-14
Respect trakt POST rate limit

twolaw:
Changed I added so far to this PR :

1. respect POST requests interval to avoid rate-limit errors (instead of going full speed and get many 429 errors)
2. when 429 error (rate limit exceeded) is raised, read and apply the Retry-After value from trakt header 
3. if 429 error is raised too much time for the same request, it aborts this request to avoid infinite loop
4.  correction of logging error message

Important : the Retry-After value can be read from response only when [this PR](https://github.com/moogar0880/PyTrakt/pull/138) of PyTrakt will be merged. So we should **wait before merging this**.

Nota: TVshow episodes are synced one by one because of the PyTrakt module structure, which is not great (read [here](https://github.com/moogar0880/PyTrakt/issues/132)) therefore trakt rate limit is reach very quickly.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,4 +1,3 @@
-
 import plexapi.server
 from os import getenv, path
 import trakt
@@ -23,6 +22,8 @@ import requests_cache
 
 requests_cache.install_cache('trakt_cache')
 env_file = path.join(path.dirname(path.abspath(__file__)), ".env")
+trakt_post_wait = 1.2  # delay in sec between trakt post requests to respect rate limit
+
 
 def process_movie_section(s, watched_set, ratings_dict, listutil, collection):
     # args: a section of plex movies, a set comprised of the trakt ids of all watched movies and a dict with key=slug and value=rating (1-10)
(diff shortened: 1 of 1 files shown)


glensc:
@twolaw i'm confused you did it this way.

I deliberately made https://github.com/Taxel/PlexTraktSync/pull/119 in taxel repository so you can push to the same branch, which you said you could not before. 

but you created another Pr that targeted that branch pr. yeah it works. but confusing.

also, how you were able to merge without approval. huh? the 1-approval applies only to target branch master?

twolaw:
> I deliberately made #119 in taxel repository so you can push to the same branch, which you said you could not before.

Yes I know. I pushed to _jbalcorn-master-pr-84_ branch but github created a PR. I was confused too. So I merged it with _jbalcorn-master-pr-84_ (no approval because not master branch I think.
Maybe there was a way to push to the branch without a formal PR... my bad.

glensc:
@twolaw you pushed to `trakt-rate-limit` branch on your repo, not taxel:

image

but anyway, interesting outcome, your commit ended up where needed. maybe it's useful to collaborate this way for bigger PR's in the future.

twolaw:
Ok I understand.
Next time I'll try like that. As you can see I'm not a git expert, I'm learning as I practice 😄

Issue #108, opened 2021-02-08
Installing requirements.txt fails on Python 3.9


flammable:
I'm using Python 3.9 on the Mac that I'm running this script (it's running macOS 10.15.7). Both pip and pip3 are pointed to `/Library/Frameworks/Python.framework/Versions/Current/bin/python3`, which at the moment is `/Library/Frameworks/Python.framework/Versions/3.9/bin/python3`.

Running `pip3 install -r requirements.txt` failed at the part where it's trying to install typed_ast, version 1.4.0. I found [this GitHub issue](https://github.com/python/typed_ast/issues/126) where it appears this was resolved in version 1.4.2. I modified requirements.txt to include `typed-ast==1.4.2` instead, and was able to install successfully. The Trakt/Plex sync works OK now.

I don't know the greater implications of this, so I figured I'd submit an issue before a pull request. If you'd like a pull request for this, just let me know.

glensc:
@flammable send PR, please. Also, you can use pipenv to run the app in isolation of system library versions.

glensc:
@flammable, @Taxel, @twolaw: do you know what for the `typed-ast` package is needed? git history is not useful here. it was added in 6831818311cc4c09b98a9c7b9886f41ecf1a1fe8 commit. when developing I removed the line from requirements, installed dependencies, and the app seems to work without errors.

flammable:
Thanks! I don't actually know what it does, just that using the newer version resolves the installation errors with Python 3.9.

I'll send a PR in a bit, but if `typed-ast` isn't needed at all, it's even better to not include it in the first place. :)

flammable:
Actually, looking at the [readme](https://github.com/python/typed_ast/blob/master/README.md)...it appears they don't plan to support `typed-ast` on Python 3.8 and greater. It doesn't look like it's necessary anymore. Maybe it should just be removed from requirements.txt instead?

glensc:
perhaps write the dependency with version range. I've seen it being done, but don't know-how.

to my knowledge, the current code in this repo does not use Python typings (but I thought of using typings), but dependencies might, and I'm not sure, do the dependencies need or use the typings. I'm just a fly-by Python coder.

glensc:
@flammable so, as all that is unknown, please send PR of you changing the requirement that worked for py3.9.

glensc:
I think what happened was that @Taxel tried to save dependencies used in the project, and used `pip freeze`:
[...]

PR #132, opened 2021-03-18
Remove bogus packages

glensc:
Keep track only of top level dependencies. Fixes https://github.com/Taxel/PlexTraktSync/issues/108

```
$ pipenv uninstall \
  astroid certifi chardet idna isort \
  lazy-object-proxy oauthlib pylint \
  requests-oauthlib six tqdm typed-ast urllib3 wrapt \
  websocket_client mccabe
```

`requirements.txt` modified manually.

See discussion at https://github.com/Taxel/PlexTraktSync/issues/108


Pipfile
```diff
@@ -6,27 +6,11 @@ verify_ssl = true
...
-certifi = "==2019.9.11"
-chardet = "==3.0.4"
-idna = "==2.8"
-isort = "==4.3.21"
-lazy-object-proxy = "==1.4.2"
-mccabe = "==0.6.1"
-oauthlib = "==3.1.0"
-pylint = "==2.4.3"
 python-dotenv = "==0.10.3"
 requests = "==2.25.0"
```
twolaw, line 10: `requests` is already a requirement for `trakt` so not needed here.
Also it's better to avoid future version conflict with trakt requirement. I sugest to remove `requests`.

Pipfile
```diff
@@ -6,27 +6,11 @@ verify_ssl = true
...
-astroid = "==2.3.2"
-certifi = "==2019.9.11"
-chardet = "==3.0.4"
-idna = "==2.8"
-isort = "==4.3.21"
-lazy-object-proxy = "==1.4.2"
-mccabe = "==0.6.1"
-oauthlib = "==3.1.0"
-pylint = "==2.4.3"
 python-dotenv = "==0.10.3"
```
twolaw, line 9: You can update to last versions of :
PlexAPI==4.4.1
python-dotenv==0.15.0

twolaw (changes requested):

Pipfile
glensc: I had impression this project used it directly, but seems only requests_cache is used directly. Will remove.

Pipfile
glensc: I do prefer keeping one change per one PR. but I guess can make an exception here.

glensc:
@twolaw changes made!

twolaw (changes requested):

requirements.txt
```diff
@@ -1,4 +1,4 @@
-PlexAPI==4.3.1
+plexapi==4.3.1
```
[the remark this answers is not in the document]
glensc: Fixed

requirements.txt
```diff
@@ -1,21 +1,4 @@
...
-chardet==3.0.4
-idna==2.8
-isort==4.3.21
-lazy-object-proxy==1.4.2
-mccabe==0.6.1
-oauthlib==3.1.0
-PlexAPI==4.3.1
-pylint==2.4.3
+plexapi==4.4.1
 python-dotenv==0.10.3
```
twolaw, line 2: 0.15.0

twolaw (changes requested):

requirements.txt
glensc: Fixed.

Now I checked all four, they are the same in Pipenv and in requirements.txt

Probably should not do multiple things in parallel :D

twolaw (approved):

PR #140, opened 2021-03-19
Rename config.json to config.defaults.json

glensc:
The `config.json` use is preserved, except it's no longer in Git, meaning users can have their own content there without having to merge on updates.

I don't consider this a breaking change, as people don't lose their changes accidentally:
1. they can't pull changes from git repo if. they have modified config.json
2. they have to commit or stash before pulling
3. and if they pull after that, they need to deal with file merge, etc

however, I'm not sure how to present this properly for users to update. Introducing a completely new path for config file would be a smooth change, in the sense that keeping the deprecated config.json in repository for some time.


glensc:
Maybe it's time to use `config/config.json` as a new config path? However, not sure what complications that brings, as if automatically migrating, need to make new dirs, which may be an issue in some environments, like the path is not writable in docker image.

or as this project is not mature at all, just go with the flow?

twolaw:
I disagree this `defaults come from code` way.
How do we customize settings before first run ?
Moreover, with a dedicated settings file (as it is right now) the user can read it before downloading this project, to check if it suits its needs.

I suggest renaming `config .json` to `config.json.default` and script copy at first run if `config .json` doesn't exist. Also explain this in readme file.
This is the way config files are setup in linux world, as far as i know.

glensc:
@twolaw but the removal of config.json from repo problem remains. how to solve that problem?

twolaw:
How is that a problem?

glensc:
@twolaw please read the PR body. it's described there.  you don't see it as a problem? If users modified their config.json they can't update smoothly anyway? so they have to manually merge removed file changes?

glensc:
Updated to use `config.default.json` instead. But this probably needs something in README?

twolaw:
This new version looks good to me. So no problem anymore, right ?

As I said :
> Also explain in readme file to copy to config.json for customized settings.

glensc:
> Also explain in readme file to copy to config.json for customized settings.

but this is not needed, it's done automatically.

glensc:
> So no problem anymore, right?

The problem of config.json removal concern I described already. not going to do that anymore.

twolaw:
> 
> 
> > Also explain in readme file to copy to config.json for customized settings.
> 
> but this is not needed, it's done automatically.

I said "before first run" :
> How do we customize settings before first run ?

The user needs to know he can copy `config.default.json` to `config.json` to setup customized settings **before first run**. That is important if user doesn't want to do a full sync.

---

> they can't pull changes from git repo if. they have modified config.json
> they have to commit or stash before pulling
> and if they pull after that, they need to deal with file merge, etc

All this is irrelevant because `config.json` is in `.gitignore`.

glensc:
Added something to the readme about config.json

README.md
```diff
@@ -67,6 +67,10 @@ To disable parts of the functionality of this software, look no further than
 `config.json`. Here, in the sync section, you can disable the following things
 by setting them from `true` to `false` in a text editor:
 
+If the `config.json` is missing, copy it from `config.default.json`, the script
+will create `config.json` on first run, so this is only needed if you intend to
+customize before first run.
+
```
twolaw, line 73: I would not start with "_if the config.json is missing..._" because the user will think something is wrong.
Here is my version :
```
At first run, the script will create `config.json` based on `config.default.json`. If you want to customize settings before first run (ex. you don't want full sync) you can copy and edit `config.json` before launching the script.
```

twolaw (changes requested):

README.md
glensc: 8481387

glensc: Note that this need may go away in the future after command-line options support.

I already myself want to have `--no-tv` or `--sync=movies` (or `--movies`) option to indicate sync only certain library type.

twolaw: Great, you had a very good idea to add this click module. Seems very promising 👍 
I'll read other PR later.

twolaw (approved):

glensc:
Finally, don't have to stash `config.json` changes when developing! I usually forget it, so I'm scanning the huge Archive library...

PR #122, opened 2021-03-06
Feature: Use click package for command line parsing

glensc:
This adds click framework support, allowing to control options from the command-line.

Adding actual commandline arguments will be submitted as future PRs.

NOTE:
- [x] Built on top of https://github.com/Taxel/PlexTraktSync/pull/121, so it needs to be merged first.

PR #141, opened 2021-03-19
Refactor: Add requests_cache module wrapper

glensc:
Extracted from https://github.com/Taxel/PlexTraktSync/pull/122 making it smaller.

### DIFF
--- a/plex_trakt_sync/main.py
+++ b/plex_trakt_sync/main.py
@@ -1,5 +1,4 @@
 import plexapi.server
-import requests_cache
 import trakt
 from plex_trakt_sync.path import pytrakt_file, env_file, trakt_cache
 from plex_trakt_sync.plex_api import PlexApi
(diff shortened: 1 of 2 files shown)


twolaw (commented):
Ok for `trakt_cache` in _path.py_ but I don't see the point of _requests_cache.py_.
Too much factoring IMO.

glensc:
Can't pull more changes from https://github.com/Taxel/PlexTraktSync/pull/122 as it would require other changes well (`@nocache` decorator), which I prefer to submit separately.

glensc:
The whole global symbols modification is a bad design, but it takes some more iterations to re-design it better:

- logging should be an instance
- CONFIG should be an instance
- this requests_cache should be an instance

I'll turn this back to Draft as merge conflicts appeared, and maybe delay this module introduction for now.

glensc:
Difficult to rebase other branches, as they depend on his change.

If this sill bothers after refactorings can submit another PR to deal with that.

PR #149, opened 2021-03-20
Update logging configuration to add stdout logger

glensc:
This makes log entries INFO and above also appear on the console. Thus simplifying code not requiring to do duplicate print and logging.


glensc:
merging to `main` to get things going:
- https://github.com/Taxel/PlexTraktSync/issues/162

PR #156, opened 2021-03-30
Test

sirloinofbeef:


data/LICENSE.md
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 Alexander Theimer
+Copyright (c) 2021 SirLoinOfBeef
```
glensc, line 3: it is a copyright violation to alter any copyright notices!

Here's a highlight from the license itself:

> **The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.**


cc @Taxel

PR #169, opened 2021-04-04
Correct connect error message on plex connection

glensc:
At the time this exception is thrown, the main url as tried, not fallback.

Found this when reading the code, no actual error encountered.

refs:
- #94 

cc @twolaw


twolaw:
Following you logic, why didn't you "correct" line 378 too ?

The error message is correct.
Your PR is wrong because you forgot to read the moment m is showed to user.

glensc:
@twolaw well, the code is just unreadable!

PR #172, opened 2021-04-05
Fix for new Plex TV agent

BEisem:
This is a fix for the new Plex TV agent.

### DIFF
--- a/plex_trakt_sync/main.py
+++ b/plex_trakt_sync/main.py
@@ -189,16 +189,21 @@ def process_show_section(s, watched_set, listutil):
     logging.info("Now working on show section {} containing {} elements".format(s.title, len(allShows)))
     for show in allShows:
         guid = show.guid
+        if guid.startswith('plex://show/'):
+            if len(show.guids) > 0:
+                logging.debug("trying first alternative guid: " + str(show.guids[0].id))
+                guid = show.guids[0].id
+        x = provider = None
         if guid.startswith('local') or 'agents.none' in guid:
             # ignore this guid, it's not matched
             logging.warning("Show [{} ({})]: GUID is local, ignoring".format(
                 show.title, show.year))
             continue
-        elif 'thetvdb' in guid:
+        elif 'imdb' in guid:
             x = guid.split('//')[1]
             x = x.split('?')[0]
-            provider = 'tvdb'
-        elif 'themoviedb' in guid:
+            provider = 'imdb'
+        elif 'themoviedb' in guid or 'tmdb' in guid:
             x = guid.split('//')[1]
             x = x.split('?')[0]
             provider = 'tmdb'


glensc:
@BEisem you forgot to add references to previous discussions in PR body. Please do that.

plex_trakt_sync/main.py
```diff
@@ -189,16 +189,21 @@ def process_show_section(s, watched_set, listutil):
     logging.info("Now working on show section {} containing {} elements".format(s.title, len(allShows)))
     for show in allShows:
         guid = show.guid
+        if guid.startswith('plex://show/'):
+            if len(show.guids) > 0:
+                logging.debug("trying first alternative guid: " + str(show.guids[0].id))
+                guid = show.guids[0].id
```
glensc, line 195: Please share what are the values of `show.guids`

plex_trakt_sync/main.py
```diff
@@ -189,16 +189,21 @@ def process_show_section(s, watched_set, listutil):
...
+        x = provider = None
         if guid.startswith('local') or 'agents.none' in guid:
             # ignore this guid, it's not matched
             logging.warning("Show [{} ({})]: GUID is local, ignoring".format(
                 show.title, show.year))
             continue
-        elif 'thetvdb' in guid:
```
glensc, line 197: why have you removed `thetvdb`? please add new scenarios, do not remove old scenaros.

BEisem: Sorry, this is literally my first time doing a pull request so I'm just fumbling through this. But my understanding was that the new TV agent no longer uses `thetvdb` which is why I removed it.

glensc: @BEisem but there's no intention to stop supporting the previous tv agents. legacy and previous agents can be removed when they can't be used, but even the latest PMS includes the legacy agents.

glensc:
closing since pr author does not respond to review feedback and this is not relevant in latest release

twolaw:
> closing since pr author does not respond to review feedback and this is not relevant in latest release

I disagree, please let people some time to polish PR for the current default branch.

plex_trakt_sync/main.py
wandabastyle: Plex metadata for the new agents can contain multiple guids, seems like imdb is the first and therefore used first in the elif-statement.

Example from my library:
`grandparentGuid="plex://show/5d9c080602391c001f57f2ce"` with guids:
```xml
<Guid id="imdb://tt5312444"/>
<Guid id="tmdb://1218828"/>
<Guid id="tvdb://5632569"/>
```
I have not seen metadata without an IMDB refenrence, but might still happen, so keeping the other loops is a good idea.

BEisem: Sorry for my ignorance, but where do you go to find this information? I would like to try and take a look at one of the episodes in my library that is causing this error, to see if I see the multiple guids that you have found, or if it is missing the IMDB reference.

BEisem: Scratch that last question, I found it in the XML of the episodes.

So I discovered that, for any episode that is throwing this error in PlextraktSync, the metadata for that episode does not contain any references to IMDB, TMDB or TVDB. It's as if the metadata is incomplete. And instead of an entry like `guid="plex://episode/5de7db195e4b67001edc92de"` it will show something like `guid="local://4036"`.

Just so we are comparing apples and apples, do you have the series "American Dad?" If you take a look at Season 15 Episode 22, this is one of the ones that is displaying this problem. In the metadata, the guid is `guid="local://4036"`. This must be what is causing the problem.

wandabastyle: I checked TMDB and there is no ep.22, therefore the plex agent might have a problem. Could be differences in numbering (tv order & dvd order and the like).

glensc: pardon me, but what error are you both talking about?

BEisem: Oops, sorry... I thought I was here: https://github.com/Taxel/PlexTraktSync/discussions/208

wandabastyle:
So how do we proceed with this pull request? I did the same changes as above, but added another loop for IMDB and I am keeping the other providers. Details I explained in #144.
I just am not sure how github collab works. Can we fix this in this request or close it and open another one?
I think I might be able to do this now. Never done it before, like @BEisem.

glensc:
@wandabastyle it should be fixed in `0.6.37` (#202 Add replace of thetvdb to tvdb):
- https://github.com/Taxel/PlexTraktSync/pull/209

wandabastyle:
@glensc should I comment here or in #202, bcs. I tested your fork and TV series with the beta agent do not work for me. I would gladly provide details if you tell me where.

glensc:
@wandabastyle #202 is merged, unless you are to be commenting on the code itself, better create a new issue of your problem. because I don't know what you did, what's your errors, how to reproduce, etc.

also, this PR author was asked the same, but they still have not filled details such:
- https://github.com/Taxel/PlexTraktSync/pull/172#issuecomment-813240435

plex_trakt_sync/main.py
BEisem: I don't know how to answer this, I'm sorry

glensc:
This fixed in 0.6.x versions, I propose we close this PR.

For merging to 0.5.x branch this PR needs to restore the previous provider names, I don't think this is worth the effort.

BEisem:
Yes, I closed it.

PR #180, opened 2021-04-05
Refactor: Sync ratings using Plex/Trakt API classes

glensc:
Movie ratings are synced now with PlexApi and TraktApi classes.

### DIFF
--- a/plex_trakt_sync/commands/sync.py
+++ b/plex_trakt_sync/commands/sync.py
@@ -38,7 +38,7 @@ def sync_all():
 
     for section in plex.movie_sections:
         with measure_time("Processing section %s" % section.title):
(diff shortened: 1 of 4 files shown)


glensc:
In my opinion, whole ratings syncing is broken.

- Plex has a float value, say `8.6`
- Trakt has an integer value (1 to 10), so `8.6` is synced as `8`
- In the next iteration, the rating is again synced even if it has not really changed.

And in fact I don't know where Plex the rating is set, visually it's only 5 stars, but the value set by API does not change that.

# Plex UI

image

# Trakt UI

image

glensc:
Got the `8.6` mistakenly because accessed mistakenly `Movie.rating` rather `PlexLibraryItem.rating`.

PR #188, opened 2021-04-06
Use our Plex/Trakt API classes to sync collections

glensc:


glensc:
Looks like previous code did not execute `m.add_to_library()` under `requests_cache.disable()`. probably harmless as caching POST method adding item to library isn't gonna get cache hit.

glensc:
The trakt web, supports including collection date

PR #219, opened 2021-04-10
Upgrade requests-cache to 0.6.x

glensc:
- [ ] previous cache will all expire on upgrade
- [x] currently (requests-cache 0.6.0) crashes loading old cache (fixed)
- [x] currently (requests-cache 0.6.0) it logs too much, needs to patch log level (fixed)

refs:
- https://github.com/reclosedev/requests-cache/issues/224
- https://github.com/reclosedev/requests-cache/issues/223

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,7 +10,7 @@ click = "==7.1.2"
 plexapi = "==4.5.0"
 python-dotenv = "==0.15.0"
 python-git-info = "==0.6.1"
-requests-cache = "==0.5.2"
+requests-cache = "==0.6.0"
 trakt = "==3.1.0"
 
 [requires]
--- a/Pipfile
+++ b/Pipfile
@@ -10,7 +10,7 @@ click = "==7.1.2"
 plexapi = "==4.5.0"
 python-dotenv = "==0.15.0"
 python-git-info = "==0.6.1"
-requests-cache = "==0.6.0"
+requests-cache = "==0.6.3"
 trakt = "==3.1.0"
 
 [requires]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "e15535215392894422f8613d1e96901b69d119d6d84c5eed61642ffc11f3df57"
+            "sha256": "a7198f235d31d57dd1775bc1c2f1017ac8ea09b5b415901c6230ac97867562de"
         },
         "pipfile-spec": 6,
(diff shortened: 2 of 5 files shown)


glensc:
maybe it's time to change the path where the cache is stored, would fix the first two problems.

i.e for docker friendliness maybe it's better to use subdirs for config and cache. 

maybe should use XDG paths for both?

glensc:
@twolaw do you have any opinion here on the pull request and the notes I've written?

twolaw:
> maybe it's time to change the path where the cache is stored, would fix the first two problems.

I understand that's not a problem anymore, so keeping cache in the same place is ok to me.
Therefore, seems ready for merge.


> i.e for docker friendliness maybe it's better to use subdirs for config and cache.
> 
> maybe should use XDG paths for both?

I agree.
I think this should be discussed in the  docker discussion.
And you are far better than me about linux things (I never heard about XDG before).

twolaw (approved):

glensc:
> previous cache will all expire on upgrade

you already said "ok", but I want to be clear;

as this update will be unable to use the existing cache, this should go to 0.8.x or to the next version 0.9.x?

I first thought that it best to be in the next version, but very likely users update without even refreshing the dependencies, so it makes no difference to what version includes the change.

also, it's possible to convert the cache database to the new version, but I do not think it's worth the effort, so I'm leaving it out.

twolaw:
> so it makes no difference to what version includes the change.

> I do not think it's worth the effort

I agree, go for it.

glensc:
I wonder if this a new problem, or not, but the script could not see newly added tv episodes. needed to delete requests cache for them to be seen.

I've checked code and the related calls for these (`.library_sections()` and `.all()`) methods are both annotated with `@nocache`, so `requests_cache` should not cache them:
- https://github.com/Taxel/PlexTraktSync/blob/896981bf6ba659a2bbb85fb3f7acf99a14ac9934/plex_trakt_sync/plex_api.py#L211-L221
-  https://github.com/Taxel/PlexTraktSync/blob/896981bf6ba659a2bbb85fb3f7acf99a14ac9934/plex_trakt_sync/plex_api.py#L156-L159

twolaw:
can memoize be the problem ?

glensc:
@twolaw no, memoize is in memory only, so it's destructed once the process exists.

I'm trying to understand the cache contents for now:
- https://github.com/Taxel/PlexTraktSync/pull/296

and compare to older versions of PlexTraktSync and requests_cache.

PR #232, opened 2021-04-12
Group API calls to add items to Trakt collection

glensc:
This reduces API calls made to Trakt API to add items to collection to 1 (or 0).

This also reports summary of changes made:
```
INFO: Updated Trakt collection: {'added': {'episodes': 15}}
INFO: Completed full sync in 45.0 seconds
```

this saved 14 POST calls to Trakt, possibly reducing run time of 13 seconds.

### DIFF
--- a/plex_trakt_sync/commands/sync.py
+++ b/plex_trakt_sync/commands/sync.py
@@ -178,6 +178,8 @@ def sync_all(movies=True, tv=True):
     with measure_time("Updated plex watchlist"):
         listutil.updatePlexLists(server)
 
+    trakt.flush()
+
 
 @click.command()
 @click.option(
--- a/plex_trakt_sync/trakt_api.py
+++ b/plex_trakt_sync/trakt_api.py
@@ -51,6 +51,9 @@ class TraktApi:
     Trakt API class abstracting common data access and dealing with requests cache.
     """
 
+    def __init__(self):
+        self.batch = TraktBatch(self)
+
     @property
     @memoize
     @nocache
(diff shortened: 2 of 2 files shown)


glensc:
> This reduces API calls made to Trakt API to add items to collection to 1 (or 0).

This may become an issue in the future if the batch size goes too big. Postponing worrying about the problem right now. Deal with the problem when it becames necessary.

One issue may be if someone's collection is too big and their initial run gets aborted and they retry and have to issue everything from the start, for that --batch-size option could do. but again, deal when it becomes necessary.

BEisem:
> > This reduces API calls made to Trakt API to add items to collection to 1 (or 0).
> 
> This may become an issue in the future if the batch size goes too big. Postponing worrying about the problem right now. Deal with the problem when it becames necessary.
> 
> One issue may be if someone's collection is too big and their initial run gets aborted and they retry and have to issue everything from the start, for that --batch-size option could do. but again, deal when it becomes necessary.

This must be what I've been running into here: https://github.com/Taxel/PlexTraktSync/issues/256#issuecomment-821851004

My library is too large to group everything together at the end. A --batch-size option would definitely be helpful in this case.

glensc:
Yeah, --batch-size is an easy addon to this.

But I as thinking, why this was done, it's to skip waiting for post requests, so the problem to solve is not to submit anything faster than 1s. But implementing that logic is rather complex, sure can see if the last submit was <1s ago, add to the batch, otherwise submit (and flush batch queue). but there's still no 'trigger' for the last item to flush. so, for that need a timer or something to check the queue at 1s intervals.

Also, probably good idea is to add a hook to flush the queue on the program exit. I'll see if `click` framework has some recipe for such "program exit" hook.

glensc:
Doesn't seem that click has any "exit hook" or suitable exception
- https://click.palletsprojects.com/en/7.x/exceptions/

however, python seems to have __del__ for destructor:
- https://www.geeksforgeeks.org/destructors-in-python/
- https://pythonprogramminglanguage.com/destructor/

however, it's called too late, when many things are already destroyed, like logger no longer works:

```
Exception ignored in: <function TraktBatch.__del__ at 0x10594da60>
Traceback (most recent call last):
  File "/Users/glen/scm/plex/Plugins/PlexTraktSync/plex_trakt_sync/trakt_api.py", line 244, in __del__
  File "/usr/local/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 1458, in warning
  File "/usr/local/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 1589, in _log
  File "/usr/local/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 1599, in handle
  File "/usr/local/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 1661, in callHandlers
  File "/usr/local/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 952, in handle
  File "/usr/local/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 1186, in emit
  File "/usr/local/Cellar/python@3.9/3.9.4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/logging/__init__.py", line 1175, in _open
NameError: name 'open' is not defined
```

So, no "magic" solution here, just need to catch top-level Exception in main app

PR #247, opened 2021-04-16
Add --show option to sync, to sync only specific show

glensc:
```
Usage:

./plex_trakt_sync.sh sync --show='The Addams Family (1964)'
```

requires:
- [x] https://github.com/Taxel/PlexTraktSync/pull/246

PR #249, opened 2021-04-16
Raise exception when --show option resulted no match

glensc:
Feature request:
- https://github.com/Taxel/PlexTraktSync/pull/247#issuecomment-821222972


glensc:
To be solved in https://github.com/Taxel/PlexTraktSync/pull/282 instead

Issue #231, opened 2021-04-12
Feature request -- allow user to determine log file name


BEisem:
It would be nice to allow the user to differentiate log files for different types of scans.

For example, I do a TV scan every couple of hours. But I only do a movie scan once a day. Since the script erases the log file and starts over each time, it is difficult for me to look at the log file for the movies, since it is getting overwritten by the latest log file from a TV scan.

I think it would be helpful to include a command line argument to determine the name of the log file. Or some other way to separate the logs so this won't happen.

glensc:
I was thinking of moving logging configuration to config.json:
1. append: bool
1. rotate: bool
2. filename: string

glensc:
logging configuration docs:
- https://docs.python.org/3/library/logging.html#logging.basicConfig

glensc:
there's also a module to load logging config from the file:
- https://docs.python.org/3/library/logging.config.html#module-logging.config

but not sure if it's a good idea to give full control of logging to end-users...

BEisem:
Sorry, this is a bit over my head. But I think rotating the logs would be fine, otherwise giving users the ability to choose a log file name on the command line (overriding the default). But rotating seems to be a good idea as well.

glensc:
I propose such schema:

```json5
{
	"logging": {
		"filename": "PlexTraktSync.log", // previously "last_update.log'
		"level": "debug" | "info", // previously: "log_debug_messages":bool
		"append": bool, // whether to truncate file on each run 
		"rotate": bool, // enable rotation based on some built in defaults
		"rotate": [ // advanced rotate config
			"max_bytes": 1024,
			"backup_count": 3,
		],
	}
}
```

for your scenario just need:
```json5
{
	"logging": {
		"append": bool, // whether to truncate file on each run 
	}
}
```

glensc:
@BEisem: no answer.

BEisem:
Sorry, yes I think this is a good plan...

glensc:
Here's the implementation: https://github.com/Taxel/PlexTraktSync/pull/253.

Respond there if it works as expected for you.

glensc:
0.7.22: #253: Add support to disable log truncation on each run

PR #253, opened 2021-04-16
Add support to disable log truncation on each run

glensc:
This solves a problem that different runs overwrite logs that user is having.

```json
    "logging": {
        "append": true
    },
```

Fixes https://github.com/Taxel/PlexTraktSync/issues/231

### DIFF
--- a/plex_trakt_sync/config.py
+++ b/plex_trakt_sync/config.py
@@ -19,26 +19,6 @@ class Config(dict):
         "TRAKT_USERNAME",
     ]
 
-    defaults = {
-        "log_debug_messages": False,
-        "logging": {
-            "append": False,
-        },
-        "sync": {
-            "liked_lists": True,
-            "watchlist": True,
-            "watched_status": True,
-            "collection": True,
-            "ratings": True,
-        },
-        "xbmc-providers": {
-            "movies": "imdb",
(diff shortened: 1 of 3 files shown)


BEisem:
This works as expected. Thank you!

plex_trakt_sync/config.py
```diff
@@ -27,6 +47,8 @@ def __getitem__(self, item):
         return dict.__getitem__(self, item)
 
     def initialize(self):
+        self.update(self.defaults)
+
         if not exists(config_file):
             with open(default_config_file, "r") as fp:
                 defaults = json.load(fp)
```
Taxel, line 54: Doesn't your addition make the default config file obsolete? If so, doesn't the `if not exists...` load the default config file and save it as the new config on the first run without a config file, overriding `self.defaults` in every future execution?

Lines 55 and 56 are also part of the behavior I am describing but I could not select them.

glensc: @Taxel this code is executed only if the config file does not exist.

But you gave me an idea of how to refactor this: must load the default config ourselves, and then user config.

glensc: Fixed in f94bc56

PR #254, opened 2021-04-16
Refactor: Add Media container

glensc:
Purpose of this class is to create unified Media object that has Plex and Trakt media items.

This will simplify code where having to pass around plex/trakt pairs, and likely avoid copy-paste errors.

### DIFF
--- a/plex_trakt_sync/commands/sync.py
+++ b/plex_trakt_sync/commands/sync.py
@@ -12,15 +12,15 @@ from plex_trakt_sync.logging import logger
 from plex_trakt_sync.version import git_version_info
 
 
-def sync_collection(pm, tm, trakt: TraktApi, trakt_movie_collection):
+def sync_collection(m: Media, trakt: TraktApi, trakt_movie_collection):
     if not CONFIG['sync']['collection']:
         return
 
-    if tm.trakt in trakt_movie_collection:
+    if m.trakt.trakt in trakt_movie_collection:
         return
 
-    logger.info(f"To be added to collection: {pm}")
-    trakt.add_to_collection(tm, pm)
+    logger.info(f"To be added to collection: : {m.plex}")
+    trakt.add_to_collection(m.trakt, m.plex)
 
 
 def sync_show_collection(tm, pe, te, trakt: TraktApi):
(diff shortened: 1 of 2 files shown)


Taxel:
Big fan of consolidating plex and trakt media items 👍 
One question: is there a reason why you named the Media instance passed as an argument `m`, but sometimes `me`? Always naming it the same would make the code a little more consistent.

glensc:
@Taxel The need for a single item became from the incremental refactoring. As I needed to pass on the context, I ended up passing 4 variables sometimes, which can lead to very simple bugs but difficult to discover.

As for naming:
- https://www.mediawiki.org/wiki/Naming_things
> "There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors." – [Leon Bambrick][1]

[1]: http://martinfowler.com/bliki/TwoHardThings.html

So, I did not have a good name, so went with something plain.

as for variable names, I'm using Python typing now, the methods are small and simple, there's no real need for descriptive variable names if the naming is consistent over the project.

The convention used (for now):
- `m: Media' of type Movie or Show
- `me: Media` of type Episode
- `pm: Plex Media`
- `tm: Trakt Media`
- `pe: Plex Episode`
- `te: Trakt Episode`
- ...

this is not carved to stone, can be changed later, but I found it easier to refactor this way.

glensc:
I did read your question again, so I'll answer your question :)

Some code, especially dealing with episode, have both variables, `m` for show and `m` for episode, so as it's bad practice to overwrite variables with a different purpose, I intrudoced new variable, `e` stands for episode.

Taxel:
Ah, thanks for clarifying that. When reading the code I mainly looked at the type and the similar names and wondered if they are the same.
On a side note: Aren't the typings a part of Python 3.5, requiring a Python version bump in the pipfile?

glensc:
Don't know about typings version, but in readme we say python 3.6 is required. And I've already written some code that requires python 3.6, so can't go lower than that, also 3.5 seems eol, so not worth the effort:
- https://endoflife.date/python

The typings aren't runtime checked, so the files just need to parse, so I did not worry much about it.

I'll have look, why Pipfile has version "3", I'm sure there was version there before.

PR #255, opened 2021-04-17
Add documentation section about how to share PMS database

glensc:


glensc:
I was thinking, maybe add instructions how to remove/obfuscate some sensitive data:
1. obfuscate account names
2. remove password hashes

deleting account entries is not a good option, as that would make the database inconsistent, so better to keep the records to have their id present in the database.

glensc:
cool, seems the downloaded database already has removed hashed passwords (the column is empty).

glensc:
And doc how to restore:
- https://support.plex.tv/articles/202485658-restore-a-database-backed-up-via-scheduled-tasks/

as the backup did not contain the blobs database, I just deleted the old blobs database.

Issue #257, opened 2021-04-17
Crash happening during movie sync


BEisem:
0.7.19 causes a crash during a movie sync.

End of the script output before the crash:

```
Processing Movies  [#################-------------------]  5268/11117  00:48:45INFO: Add to Trakt Collection: <imdb:tt11908982:<Movie:198327:Jakob's-Wife>>
Processing Movies  [#################-------------------]  5275/11117  00:48:41INFO: Add to Trakt Collection: <imdb:tt0097613:<Movie:197966:The-January-Man>>
Processing Movies  [#################-------------------]  5404/11117  00:47:33INFO: Add to Trakt Collection: <imdb:tt0045943:<Movie:197979:Julius-Caesar>>
Processing Movies  [#################-------------------]  5422/11117  00:47:24INFO: Add to Trakt Collection: <imdb:tt6710826:<Movie:198267:The-Jurassic-Games>>
Processing Movies  [#################-------------------]  5453/11117  00:47:09
Traceback (most recent call last):
  File "./main.py", line 6, in <module>
    cli()
  File "/Library/Python/3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Library/Python/3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Library/Python/3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Python/3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Library/Python/3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/bink/scripts/PlexTraktSync/plex_trakt_sync/commands/sync.py", line 235, in sync
    sync_all(movies=movies, tv=tv, show=show)
  File "/Users/bink/scripts/PlexTraktSync/plex_trakt_sync/commands/sync.py", line 178, in sync_all
    for pm, tm in for_each_pair(plex.movie_sections, trakt):
  File "/Users/bink/scripts/PlexTraktSync/plex_trakt_sync/commands/sync.py", line 116, in for_each_pair
    tm = trakt.find_by_media(pm)
  File "/Users/bink/scripts/PlexTraktSync/plex_trakt_sync/decorators/rate_limit.py", line 45, in wrapper
[...]

PR #248, opened 2021-04-16
Merge up 0.7.x to main

glensc:
Similar to 0.6.x: https://github.com/Taxel/PlexTraktSync/pull/209

Changes:
- [0.7.22]: #253: Add support to disable log truncation on each run
- [0.7.21]: #251: Update find_by_media to support legacy episode format
- [0.7.20]: #250: Add trakt.find_episode method find Trakt episode from Plex episode
- [0.7.19]: #239: Update rate_limit to also catch ConnectionError
- [0.7.18]: #230: Add command to log in to Plex
- [0.7.17]: #243: Add trakt login command
- [0.7.16]: #242: Update Pytrakt 3.1.0
- [0.7.15]: #247: Add --show option to sync, to sync only specific show
- [0.7.14]: #246: Add search API to Plex
- [0.7.13]: #244: Add semantic styles helpers
- [0.7.12]: #241: Cleanup code to yield te.instance directly rather resolve it later
- [0.7.11]: #238: Swap media_type/type names
- [0.7.10]: #237: Add inspect command
- [0.7.9]: #232: Group API calls to add items to Trakt collection
- [0.7.8]: #233: Use Plex collected_at for Episodes
- [0.7.7]: #215: Add test for episode lookup
- [0.7.6]: #225: Simplify code by using Plex episode object
- [0.7.5]: #224: Update progressbar to be more responsive
- [0.7.4]: Merge from 0.6.x branch:
  - #223: Add error catching of NotFound plex items
  - #222: Fix TV episodes matching for new TV agent
  - #221: Fix trakt username display on startup
- [0.7.3]: #218: Use property type to check media type
- [0.7.2]: #212: Show progressbar when processing section
- [0.7.1]: #213: Restart websocket if connection was closed
- [0.7.0]: #211: Use Plex collected_at for Movies

[0.7.22]: https://github.com/Taxel/PlexTraktSync/releases/tag/0.7.22
[0.7.21]: https://github.com/Taxel/PlexTraktSync/releases/tag/0.7.21
[0.7.20]: https://github.com/Taxel/PlexTraktSync/releases/tag/0.7.20
[0.7.19]: https://github.com/Taxel/PlexTraktSync/releases/tag/0.7.19
[0.7.18]: https://github.com/Taxel/PlexTraktSync/releases/tag/0.7.18
[0.7.17]: https://github.com/Taxel/PlexTraktSync/releases/tag/0.7.17
[...]

PR #258, opened 2021-04-17
Catch TraktException in rate limit to retry after 1s

glensc:
This is quick fix for https://github.com/Taxel/PlexTraktSync/issues/257

Similarly to #248, just add another exception to the catch.

refs:
- https://github.com/Taxel/PlexTraktSync/issues/260


glensc:
The exception will be caught and retried after 1s for 5 times, and the still aborted.

But there's nothing that can be done if Trakt is having server issues other than stop processing.

Issue #256, opened 2021-04-17
Movies are no longer being added to Trakt collections


BEisem:
I just noticed something ... for the last 2 days, the scripts has stopped adding movies to my Trakt collection, even though I'm running it every night. I'm currently going back and testing each version of the script, to see where the problem started ... it will display for example:

```
Processing Movies  [#-----------------------------------]  372/11114  01:19:57
INFO: Add to Trakt Collection: <imdb:tt9562694:<Movie:379251:Alien-Warfare>>
```

But if I look at the Trakt website, that movie is not added to my collection. And if I run the script again, it displays the same thing.

I'm currently working my way back through the versions to see when this started. Right now I'm testing 0.7.17. If you have any idea when code was changed that could have caused this, let me know so I can test it... otherwise, I'll report back with the latest version that it worked and the first version where it was broken.

BEisem:
Ok, found it. Something happened between 0.7.8 and 0.7.9. 

Using 0.7.8, `./plex_trakt_sync.sh sync --sync=movies` will report that a movie is being added to the Trakt collection, and it actually adds it (checked on the Trakt website). 

With 0.7.9 and higher, it reports that the movie has been added to the Trakt collection, but it doesn't actually add it. Also, repeated runs will say that it is adding the movie again.

I think we were working on TV collections when this version was created, so something must have been changed that broke the movie scan.

glensc:
the diff:
- https://github.com/Taxel/PlexTraktSync/compare/0.7.8...0.7.9

this includes hs pr:
- https://github.com/Taxel/PlexTraktSync/pull/232

it added batching the requests. please show the end of the script, there should be line about batch submission

BEisem:
I've actually been cancelling it after the first movie gets added. It would take an hour or so to run through the entire movie library. Which version do you want me to run it with? 0.7.9? I can run it and let the entire thing finish, I just need to know which version to use.

BEisem:
It looks like the same thing might be happening for TV shows as well. I am getting some shows added to the Trakt collections, but I'm wondering if it's not actually being done by the script.

Running the TV library right now, and it is claiming some episodes have been added, but when I check in the Trakt website, they are not added.

I will let the script finish (for TV shows) and see what happens at the end.

BEisem:
[...]

PR #259, opened 2021-04-17
Add --batch-size option for TraktBatch

glensc:
```
Usage: main.py sync [OPTIONS]

  Perform sync between Plex and Trakt

Options:
  --show TEXT             Sync specific show only
  --sync [all|movies|tv]  Specify what to sync  [default: all]
  --batch-size INTEGER    Batch size for collection submit queue  [default: 1]
  --help                  Show this message and exit.
```

refs:
- https://github.com/Taxel/PlexTraktSync/pull/232#issuecomment-821866092
- https://github.com/Taxel/PlexTraktSync/issues/256#issuecomment-821851004

### DIFF
--- a/plex_trakt_sync/commands/sync.py
+++ b/plex_trakt_sync/commands/sync.py
@@ -219,8 +219,8 @@ def sync_all(movies=True, tv=True, show=None, batch_size=None):
 @click.option(
     "--batch-size", "batch_size",
     type=int,
-    default=None,
-    help="Specify batch size for collection submit queue"
+    default=1, show_default=True,
+    help="Batch size for collection submit queue"
 )
 def sync(sync_option: str, show: str, batch_size: int):
     """
--- a/plex_trakt_sync/commands/sync.py
+++ b/plex_trakt_sync/commands/sync.py
@@ -154,12 +154,12 @@ def for_each_show_episode(pm, tm, trakt: TraktApi):
         yield tm, pe, te
 
 
-def sync_all(movies=True, tv=True, show=None):
+def sync_all(movies=True, tv=True, show=None, batch_size=None):
     with requests_cache.disabled():
         server = get_plex_server()
     listutil = TraktListUtil()
     plex = PlexApi(server)
-    trakt = TraktApi()
+    trakt = TraktApi(batch_size=batch_size)
 
     with measure_time("Loaded Trakt lists"):
         trakt_watched_movies = trakt.watched_movies
@@ -216,7 +216,13 @@ def sync_all(movies=True, tv=True, show=None):
     default="all",
     show_default=True, help="Specify what to sync"
 )
-def sync(sync_option: str, show: str):
+@click.option(
+    "--batch-size", "batch_size",
+    type=int,
+    default=None,
+    help="Specify batch size for collection submit queue"
+)
+def sync(sync_option: str, show: str, batch_size: int):
     """
     Perform sync between Plex and Trakt
     """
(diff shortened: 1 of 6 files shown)


glensc:
@BEisem you can use --batch-size=1 to submit queue immediately.

I have not tested at all this code, will test later. But if you wish you can test and leave your feedback.

twolaw:
In help command can you tell the unit of batch size ? is it kB ?

glensc:
@twolaw it is amount.

glensc:
@twolaw please help to rephrase it. in my opinion, it's clear already: "Specify batch size for collection submit queue"

BEisem:
Ok so if I set it to `--batch-size=10` it will collect 10 items before sending it to Trakt? That works... (haven't had a chance to test this yet, will do so shortly)

twolaw:
Ok why would the average user need to access this variable and change it? and how could one know what a good batch-size value is ?
I may be wrong but I think this is internal setting just like trakt_post_delay.

BEisem:
Actually I like the idea of having the option available to users. This way each user can choose their own "risk/reward" ... risk of losing entries with a higher amount, but fewer API calls... and lower risk with a lower number, but at the expense of higher API usage.

Maybe whatever is decided to be the best option can be the default, and it will only change from that if the user adds `-batch-size=` in the config file.

I set it to `batch-size=10` and I'm currently going through my movie library ... it's the first time in 2 days that I'm getting entries in my Trakt account, since previously it was crashing and not making it all the way through. I also like that it shows a status update every time it adds to the collection: `INFO: Updated Trakt collection: {'added': {'movies': 10}}`

I still think "Add to Trakt Collection:" should be changed to something like "Added to queue" or "Added to batch" to be more clear. But this is a great step in the right direction!

twolaw:
> Actually I like the idea of having the option available to users. This way each user can choose their own "risk/reward" ... risk of losing entries with a higher amount, but fewer API calls... and lower risk with a lower number, but at the expense of higher API usage.

I disagree. IMHO this should not be in user settings.
You spent time to read the code and understand it so this setting makes sense to you.
But do not forget the [average user](https://en.wikipedia.org/wiki/KISS_principle) who will not understand this risk/reward thing about API calls and would say : "_I don't care, just make the sync work_"

----

I agree with you about the logs.
And another solution is to log all items of the batch only when the batch is successfully sent : "Add to Trakt Collection: <Movie1> <Movie2> <Movie3> <Movie4> <Movie5>"

glensc:
`--batch-size` option is not for average users. It's for advanced users who understand what is batching and what would it improve.

And the default should be `1`, so by default the submit would be slow, but accurate.

The solution would be to make batching logic smarter:
- https://github.com/Taxel/PlexTraktSync/pull/232#issuecomment-821870284

glensc:
linguistically `Add to Trakt Collection` is correct, as it doesn't say "Added", but "Add", which means adding, but doesn't say "when" :D

When the batch is submitted, there is a message, it cames from Trakt itself, so I would leave it as is:
- https://github.com/Taxel/PlexTraktSync/pull/232

Changing the other message for now to `To be added to collection`.

twolaw:
> The solution would be to make batching logic smarter
> 

Here is a smart solution :
By default, it should flush the batch on a time base. Let's say the trakt limit is 1sec between each, we add 0.5sec as a security delay and send batch every 1.5sec
If trakt changes its limit in the future, the script will adapt automatically.
It is better than sending 1 by 1 with a useless 1.1s delay.

glensc:
@twolaw if you had clicked the link, it described there the exact same logic.

glensc:
also, do note that the default POST delay does not bother typical use cass. i.e when you run sync periodically, you add 1-2 movies shows at time. and it's not that much to sleep some seconds between them. it cames relevant if you clear your trakt collection, or run the script for the first time.

glensc:
I'm going to merge this as it is now. It restores previous behavior (before 0.7.9 #232), and if someone wants faster initial sync. they can now use --batch-size=100 or the like.

PR #263, opened 2021-04-18
Setup pyproject using setup.cfg

glensc:
Implement https://github.com/Taxel/PlexTraktSync/issues/414

The goal is to make the package installable with `pip`, maybe even runnable.

To build the sdist/wheel (needs [workaround](https://github.com/s0undt3ch/setuptools-declarative-requirements/issues/5#issuecomment-894673665)):
```sh
python3 -m pip install --upgrade build
python3 -m build
pip3 install dist/PlexTraktSync-0.13.x-py3-none-any.whl
```

without wheel (works):
```
python3 -m pip install --upgrade build
python3 -m build
pip3 install dist/PlexTraktSync-0.13.x.tar.gz
```

refs:
- https://packaging.python.org/tutorials/packaging-projects/
- https://setuptools.readthedocs.io/en/latest/setuptools.html#metadata
- https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html
- https://packaging.python.org/guides/using-manifest-in/#using-manifest-in
- https://github.com/speijnik/plex-watched-sync/blob/v0.1.1/pyproject.toml
- https://github.com/carlosperate/awesome-pyproject


glensc:
Managed to install source files. omitted the source dir override for them to be found.

Now missing is the packaging of config.default.json and entry point. May need some refactoring so they are placed under module dir.

glensc:
https://github.com/Taxel/PlexTraktSync/pull/428 makes `main.py` obsolete, so one less problem to solve.

glensc:
this adds `__version__` into `plex_trakt_sync/__init__.py`.

what I don't like it's being manually maintained. meaning it should be updated before each release.

perhaps the pain will go away if github actions is used to maintain it.

glensc:
Actually, with #461 merged, can dynamically change version of the built package:

```
$ APP_VERSION=$(git describe --tags) python3 -m build
[...]
Successfully built PlexTraktSync-0.13.19-12-g296b1cd.tar.gz and PlexTraktSync-0.13.19_12_g296b1cd-py3-none-any.whl
```

PR #282, opened 2021-04-19
Add Walker class

glensc:
Prototype for the class that can be configured as what to find and walk.

This refactors filtering to Waker class, making sync command to do just;

```python

    for movie in walker.find_movies():
        print(movie)

    for episode in walker.find_episodes():
        print(episode)
```

### DIFF
--- a/plex_trakt_sync/commands/sync.py
+++ b/plex_trakt_sync/commands/sync.py
@@ -1,4 +1,5 @@
 import click
+from plexapi.server import PlexServer
 
 from plex_trakt_sync.media import MediaFactory, Media
 from plex_trakt_sync.requests_cache import requests_cache
@@ -10,6 +11,7 @@ from plex_trakt_sync.trakt_api import TraktApi
 from plex_trakt_sync.trakt_list_util import TraktListUtil
 from plex_trakt_sync.logging import logger
 from plex_trakt_sync.version import git_version_info
+from plex_trakt_sync.walker import Walker
 
 
 def sync_collection(m: Media):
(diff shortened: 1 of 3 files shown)


glensc:
@Taxel do you have idea or suggestion or vision how to solve this problem (these problems):

users want to walk their library with different inputs;
1. all libraries of type movies or shows (default)
2. libraries of type movies (usually to split their runs)
1. libraries of type shows (usually to split their runs)
2. specific movie (usually to debug a problem)
3. specific tv (usually to debug a problem)

some of the settings come from config (excluded-libraries),
other options come from the command line, some of the options can be repeated (like --show and --movie).

as the current code ([0.8.5](https://github.com/Taxel/PlexTraktSync/releases/tag/0.8.5)) is rather messy already, even it supports only some of the wanted features, so decided some class like this is needed.

was thinking to use `yield` and `yield from` to solve the problem.

yield does have a problem, it messes up progress bar, as progress bar needs to know the value of 100%.
of course, can flatten generators with `list(generator)`, but that makes the ux jerky, long wait before progressbar can be started.

also, there are options that conflict with each other, or override, need to put that logic in too. like if you specify --library, then --show and --movie options become irrelevant.

as you've shown some interest in some of the huge refactoring pull requests (like the media object), though to ask your input.

glensc:
The walker is now feature complete and integrated to the sync command.

However, I'm not sure how to integrate progress bar :(

glensc:
Integrated progressbar into Walker, as the iterable and label and length are very tightly integrated.

glensc:
Now complete, ready for merge. Implements the same functionality as last 0.8.x.

Next steps should be to lay grounds for how the filtering actually works.

PR #301, opened 2021-04-27
Remove slow server info printing

glensc:
Removes printing of:
1. <del>server version</del>
2. last added media

those are not useful, wastes time on roundtrip for the requests, wastes resources on the server to calculate the result, and slows down the script startup.

for me, those take off 1-3 seconds of startup time.

if such information is required for debugging, we can add specific command for that.

### DIFF
--- a/plex_trakt_sync/commands/sync.py
+++ b/plex_trakt_sync/commands/sync.py
@@ -75,7 +75,6 @@ def sync_all(walker: Walker, trakt: TraktApi, server: PlexServer):
 
     with requests_cache.disabled():
         logger.info("Server version {} updated at: {}".format(server.version, server.updatedAt))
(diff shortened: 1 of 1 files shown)


twolaw:
I agree, **Recently added** is useless.
For **Server version**, I would make it run only if log_debug_messages is set to True in config file.

glensc:
I'll just remove it for now. can add when someone requests it. and I think it belongs to `--version` or `--about`.

twolaw:
Well I'm asking righ now 😄 

Why don't you want to add this simple line :
`if CONFIG['log_debug_messages']:`

It would be helpfull to help non tech-savvy users, all usefull info in one place.

glensc:
I just don't want to. So I'm removing that change now, not to deal with the problem.

Issue #307, opened 2021-04-29
0.9.x caches Plex requests


glensc:
### The problem

Plex requests are cached, so new library items are not picked up by the script.

### Expected behavior

Plex requests should not be cached.

### Steps to reproduce the behavior

```
# run the sync
./plex_trakt_sync.sh

# inspect cache, there should not be plex urls
./plex_trakt_sync.sh cache --limit=0
```

### Workarounds

Use `0.8.x` versions.

### Environment

- Version: `0.9.x`

glensc:
my guess is that either `@nocache` doesn't support nesting, or the requests are made by some other method that does not have `@nocache` decrorator.

and I think the problem appeared already with first PR merged for 0.9.x branch:
- https://github.com/Taxel/PlexTraktSync/pull/282

but it doesn't make much sense, it doesn't touch such code parts...

glensc:
So, this was the last problem that bothered me on 0.9.x branch, so pushed 0.9.x releases out and created merge up:
- https://github.com/Taxel/PlexTraktSync/pull/310

glensc:
Oh yes, my suspicion was exactly correct:
- https://github.com/pkkid/python-plexapi/blob/4.5.2/plexapi/server.py#L107

```py
    def __init__(self, baseurl=None, token=None, session=None, timeout=None):
...
        self._session = session or requests.Session()
```

it saves a copy of the requests object, so therefore `requests_cache` can't do monkey-patch on/off.

PR #308, opened 2021-04-29
Fix Plex requests being cached

glensc:
Add @nocache to get_plex_server method, Add wrapper class as `@nocache` requires class methods.

The breakage came form this being get_plex_server being not called with request cache disabled:

```py
    with requests_cache.disabled():
        server = get_plex_server()
```

Somehow that affected also other calls made by `server` later on.

Fixes https://github.com/Taxel/PlexTraktSync/issues/307


glensc:
I can't explain to myself why this change fixes problem. Does not match with my understanding how the request_cache works. 

The only logical explanation would be if Plex API makes a copy of the requests object to the plex server object...

PR #311, opened 2021-04-29
Update requests-cache to 0.7.0

glensc:
see https://github.com/reclosedev/requests-cache/issues/251#issuecomment-829588543


glensc:
This does make the read access faster. so let's go ahead with the pre-release.

PR #320, opened 2021-05-30
include commands and decorators folders as .dockerignore exclusions

Matty666:
It seems that the py files in the subfolders are being ignored and not being copied to the image.

the globstar syntax `!/plex_trakt_sync/**/*.py`, and even single folder wildcard syntax `!/plex_trakt_sync/*/*.py` don't seem to work for me either, so I had to specify the folder exclusions individually.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,5 @@
 !/Pipfile.lock
 !/config.default.json
 !/plex_trakt_sync/*.py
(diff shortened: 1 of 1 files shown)


glensc:
yeah, the `.dockerignore` globber is pretty dumb. and then they have multiple implementations for it, i.e docker, docker-compose, they implement without any code reuse.

Issue #312, opened 2021-05-06
Getting "WARNING: Can't use the AlertListener without websocket"


vladola:
### The problem

Getting WARNING: Can't use the AlertListener without websocket

### Expected behavior

It should listen for me watching stuff on plex

### Steps to reproduce the behavior

Simply run the script with "watch" argument

### Workarounds

no

### Environment

- Version: latest release
- Python version: 3.9
- Platform: e.g. `Ubuntu 20`

glensc:
@vladola please complete the issue:
1. what app version
2. complete reproducer

five2seven:
Same issue here. Using version 0.9.15. 
Input: `./plex_trakt_sync.sh watch`
Output: `WARNING: Can't use the AlertListener without websocket`

glensc:
@five2seven does `pip3 inbstall websocket` help?

xsnaruto:
> pip3 inbstall websocket

It helps to me. I didn’t realize that i dont have this module. 

But I don’t know why I can’t install websocket with pip, maybe because some issues of python versions. Anyway, at the end, I have installed it by `sudo apt install python3-websocket`.

Btw, my hardware is Raspberry 4B, Debian10.

glensc:
> But I don’t know why I can’t install websocket with pip, maybe because some issues of python versions. (I don’t know really know python).

Without sharing your errors, nobody else will know either. and it's probably best to be reported to websocket package issue system.

xsnaruto:
> > But I don’t know why I can’t install websocket with pip, maybe because some issues of python versions. (I don’t know really know python).
> 
> 
> 
> Without sharing your errors, nobody else will know either. and it's probably best to be reported to websocket package issue system.

I already finished install python-websocket with apt and I believe that's just because my own problems(python2 and python3). 

So, I didn't put my logs on. Just to confirm that why I had the same problem as this issue is I don't have python-websocket install on my machine.

PR #323, opened 2021-06-04
Add websocket-client 1.0.1 dependency

glensc:
The `websocket` package is required to run in "watch" mode:
- https://github.com/Taxel/PlexTraktSync/issues/312

Fixes https://github.com/Taxel/PlexTraktSync/issues/312

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,6 +12,7 @@ python-dotenv = "==0.15.0"
 python-git-info = "==0.6.1"
 requests-cache = "==0.6.3"
 trakt = "==3.1.0"
+websocket-client = "==1.0.1"
 
 [requires]
 python_version = "3"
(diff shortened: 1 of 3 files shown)


glensc:
This adds 5 new dependencies, not sure why it worked for me without installing this dependency, probably pipenv falls back to system packages.

we could say "to use this feature run pip3 install websocket", but then we don't control versions that get installed...

@twolaw wdyt?

twolaw:
I think `websocket` 0.2.1 is not the good one.
I think `websocket-client` is the one (maybe system package as you said because I have it on my system and the `main.py watch` command is working good). And it is an up to date project.

glensc:
Ah, my bad that, I didn't test this at all. should remove all system websocket packages to test. or use docker.

glensc:
@twolaw updated, you can merge if you have tested. I'm running off to FNS now.

twolaw (approved):

glensc:
@twolaw did you test this too besides giving approve?

PR #325, opened 2021-06-05
Call new login commands in case tokens are missing

glensc:
This replaces need for `get_env_data` code.

- adds new `login` command. used internally, but users can invoke too
- calls `trakt-login` and `plex-login` commands if respective tokens are missing

The `get_env_data` along with the `get_env_data.py` will be removed in as separate PR.


glensc:
merging, since https://github.com/Taxel/PlexTraktSync/pull/314 introduced regression writing `.pytrakt.json` to `$HOME` that this PR resolves.

glensc:
There's another problem with the click commands. If they are invoked from another command, they exit the whole program, rather returning control to original command that called.

PR #328, opened 2021-06-05
Add ensure_login method without click decorator

glensc:
This solves the problem that adding `click` decorator changed the behavior of the method.

When the first problem of click calling `sys.exit` was fixed with `standalone=False`, then next problem is that options are passed to the wrong command, i.e to `login` command, not to `sync` command.


glensc:
This removes click hacks that caused unknown options regression from https://github.com/Taxel/PlexTraktSync/pull/325

Issue #318, opened 2021-05-27
Internal Server Error for plex_trakt_sync.trakt_api.search_by_id()


CMBoii:
### The problem

_A description of what the bug is, including a complete traceback (if applicable)_

Trakt will display this issue at the very beginning but then it will scrobble the data successfully

### Expected behavior

_A description of what you expected to happen_

### Steps to reproduce the behavior

_With a complete code example, if possible_

It doesn't matter If I launch the code manually or via a cron job. 

### Workarounds

_Is there an existing workaround for this issue?_

python3 main.py

### Environment

- Version: e.g. `0.7.0` or `main` main
- Python version: e.g. `3.9` 3.8.6
- Platform: e.g. `Debian 10` DSM 6.2.4 (Synology)

```
2021-05-27 17:00:02,952 INFO[PlexTraktSync]:Syncing with Plex - and Trakt 
2021-05-27 17:00:06,623 INFO[PlexTraktSync]:Loaded Trakt lists in 3.7 seconds
2021-05-27 17:00:06,624 INFO[PlexTraktSync]:Downloaded List Trakt Watchlist
2021-05-27 17:00:06,624 INFO[PlexTraktSync]:Server version 1.23.1.4571-6119e8eed updated at: 2021-05-27 16:50:26
2021-05-27 17:00:17,329 WARNING[PlexTraktSync]:Internal Server Error for plex_trakt_sync.trakt_api.search_by_id(), retrying after 1 seconds (try: 1/5)
2021-05-27 17:00:18,847 WARNING[PlexTraktSync]:Internal Server Error for plex_trakt_sync.trakt_api.search_by_id(), retrying after 2 seconds (try: 2/5)
2021-05-27 17:00:21,381 WARNING[PlexTraktSync]:Internal Server Error for plex_trakt_sync.trakt_api.search_by_id(), retrying after 3 seconds (try: 3/5)
2021-05-27 17:00:24,909 WARNING[PlexTraktSync]:Internal Server Error for plex_trakt_sync.trakt_api.search_by_id(), retrying after 4 seconds (try: 4/5)
2021-05-27 17:00:29,437 WARNING[PlexTraktSync]:Internal Server Error for plex_trakt_sync.trakt_api.search_by_id(), retrying after 5 seconds (try: 5/5)
[...]

PR #330, opened 2021-06-06
Search all guids for a match

glensc:
**IMPORTANT**: This changeset bumps Python minimum version to 3.7:
- https://github.com/Taxel/PlexTraktSync/pull/336

NOTE: searching multiple guids requires to use of the new agent.

Future todo: 
- cleanup old unused methods
- cleanup proxy methods to guid on PlexMediaItem

refs:
- https://github.com/Taxel/PlexTraktSync/issues/318

### DIFF
--- a/plex_trakt_sync/commands/webhook.py
+++ b/plex_trakt_sync/commands/webhook.py
@@ -39,7 +39,7 @@ class WebhookHandler:
         if not plex:
             return None
 
-        media = self.mf.resolve(plex)
+        media = self.mf.resolve_any(plex)
         if not media:
             return None
 
--- a/plex_trakt_sync/media.py
+++ b/plex_trakt_sync/media.py
@@ -96,6 +96,12 @@ class MediaFactory:
         self.trakt = trakt
 
     def resolve_any(self, pm: PlexLibraryItem, tm=None):
+        # quick search for old code (faster)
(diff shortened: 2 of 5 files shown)


glensc:
I've slightly tested this locally with a legacy agent.

I've also noticed movies scan went slower: 6.2s -> 1m 38.7s, shows: 8.1s -> 13.6s, this is after eda9103 commit.

My guess is that now it has to parse all guids, even if needs one guid.

glensc:
The slowness was mitigated with 8143ca3, but I don't like having support old code too, makes codebase complicated.

glensc:
@twolaw so when I request a review, you don't even beep for 4 days.

glensc:
I like the GUID abstraction here, as appears most of the logic we do is on GUID value rather on the full Plex Media Item.

But I don't know why the new code makes things slower. probably iterating over .guids rather than reading single .guid makes more plex requests that slows things down. should compare log file with debug enabled to confirm.

Currently, we don't cache plex requests at all. We should look into doing that. I've done some research, and some refactoring needed again:
- https://github.com/reclosedev/requests-cache/issues/260

also, I think we should not build too much logic due to trakt.tv being broken, but rather help trakt.tv to fix their issues and data.

twolaw (approved):

Issue #334, opened 2021-06-12
Crash on fresh install : requests_cache AttributeError : 'HTTPResponse' object has no attribute 'isclosed'


Davst:
### The problem
Script not working, tried both with pipenv and without.

Without virtualenv:
```
Login to MyPlex was successful!
Traceback (most recent call last):
  File "main.py", line 6, in <module>
    cli()
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 1236, in invoke
    return Command.invoke(self, ctx)
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/haddoq/PlexTraktSync/plex_trakt_sync/cli.py", line 21, in cli
    sync()
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/haddoq/PlexTraktSync/plex_trakt_sync/commands/sync.py", line 136, in sync
    ensure_login()
  File "/home/haddoq/PlexTraktSync/plex_trakt_sync/commands/login.py", line 9, in ensure_login
    plex_login()
  File "/home/haddoq/.local/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
[...]

PR #337, opened 2021-06-13
Add requests and urllib3 to versioned dependencies

glensc:
Fixes https://github.com/Taxel/PlexTraktSync/issues/334

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,8 +10,10 @@ click = "==7.1.2"
 plexapi = "==4.5.0"
 python-dotenv = "==0.15.0"
 python-git-info = "==0.7.1"
+requests = ">=2.25.1"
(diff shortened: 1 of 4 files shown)


twolaw (approved):

twolaw:
Other dependencies could be `>=` instead of `==` ?

glensc:
@twolaw they could be, but so far people don't complain.

easier to support with locked versions, as less variations to support. but we can just bump the minimum, if needed.

so I'm ok to accept a PR with an open version range.

glensc:
And having fixed versions makes it difficult for people to use system/global packages, i.e installed with `pip install`.

This does not apply to pipenv, as it uses virtualenv.

PR #346, opened 2021-06-19
Setup .envrc for direnv

glensc:
https://direnv.net/


glensc:
developr docs update needed

PR #347, opened 2021-06-19
Add @deprecated decorator

glensc:
The deprecation messages can be made visible with `PYTHONWARNINGS` variable:

- https://docs.python.org/3/library/devmode.html#devmode

To print them:
```
export PYTHONWARNINGS=default
```

To make the program abort:
```
export PYTHONWARNINGS=error
```

you can set this value in `.envrc.local`:
- https://github.com/Taxel/PlexTraktSync/pull/346

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,6 +7,7 @@ verify_ssl = true
 
 [packages]
 click = "==7.1.2"
+deprecated = "==1.2.12"
 plexapi = "==4.5.0"
 python-dotenv = "==0.15.0"
 python-git-info = "==0.7.1"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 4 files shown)


glensc:
I'm confused about some things:
1. should I use external dependency?
2. should the external dependency be in requirements.txt?
3. should the external dependency be in Pipfile?

for code not to crash need to implement fallback if the import fails, whoever I planned to just do `pass` there, but the actual implementation became rather simple. I don't see what the external module does better than the local implementation.

glensc:
Also, I got hit with this problem:
- https://github.com/tantale/deprecated/issues/47

so, using a library must be a good thing.

twolaw:
Sorry @glensc it's too complicated for me and understanding will require much time i don't have 😄 
Keep going on you own, thanks a lot for the good work. I keep an eye on this project and help support in issues when I can, but will not spend much time on it.

PR #358, opened 2021-06-23
Speedup .guids access for legacy agent

glensc:
return early in guids method, if legacy agent as accessing .guids for legacy agent
will make another round-trip to plex server and the result is always empty anyway.


glensc:
This slipped in a bug of is legacy agent being always true:
- https://github.com/Taxel/PlexTraktSync/pull/360

PR #386, opened 2021-07-17
Tqdm progressbar

### LINKED ISSUE #377: Warnings not on separate line
### The problem
Some warnings are appended to the processing line rather than entered as a separate lien above.
image

### Expected behavior

Each warning displayed on a sperate line, with the processing lien beneath them.

Run sync as per normal.

### Workarounds

No

### Environment

- Version: 0.13.2
- Python version: 3.9.1
- Platform: Windows 10

mcdallas:
I managed to fix #377 by using tqdm and some minimal changes. tqdm is even [recommended by click](https://click.palletsprojects.com/en/8.0.x/utils/#showing-progress-bars) if you want a more capable progress bar. It is also much faster.

> Note
If you find that you have requirements beyond what Click’s progress bar supports, try using tqdm.

The log messages are now printed in a different line without interrupting the progress bar:

image

Fixes #377

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,6 +16,7 @@ requests-cache = "==0.7.1"
 trakt = "==3.2.0"
 urllib3 = ">=1.26.6"
 websocket-client = "==1.0.1"
+tqdm = "==4.61.2"
 
 [requires]
 python_version = "3"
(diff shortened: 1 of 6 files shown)


mcdallas:
Also note that tqdm has no upstream dependencies, it’s a pure python lib

twolaw:
Don't forget to add it in Pipfile and Pipfile.lock.

glensc:
Regression fix: https://github.com/Taxel/PlexTraktSync/pull/416

PR #395, opened 2021-07-25
add dependabot for github-actions, pip and docker

chrillep:
https://github.com/Taxel/PlexTraktSync/network/updates


glensc:
No GitHub Actions in this project, but ok to merge regardless.

chrillep:
im was thinking on adding  a workflow for tests =)
something like https://github.com/pkkid/python-plexapi/blob/master/.github/workflows/ci.yaml
as you need a a local pms for the tests :)

Issue #387, opened 2021-07-18
Cache Preventing TV Show Watched Sync


n810K:
### The problem

I am currently working through a show. I have the sync set to run every night at 3:00AM. The sync does not appear to properly mark watched episodes on Trakt. Even after manually running the sync, the episodes do not get marked as "watched". The solution that I've found is to delete the cache file then run the sync again for the episodes to be properly marked.

### Expected behavior

Episodes within a show do not properly get marked as watched

### Steps to reproduce the behavior

1) Watch an episode
2) Run the script -> Notice that the episode does not get marked on Trakt
3) Delete cache file then run script again
4) Longer run time but the episode is now properly marked as watched.

### Workarounds

1) Delete cache file then run script again
2) Longer run time but the episode is now properly marked as watched.

### Environment

- Version: 0.13.3
- Python version: 3.8
- Platform: Windows

rfgamaral:
I have a feeling this is somewhat related to #384.

glensc:
Which agent are you using?

flammable:
I'm seeing this with both movies and TV shows. In both cases, deleting the cache allows the sync to proceed.

I'm using the new movie and TV agents in Plex. I'm using version 0.13.3 of the script, Python version 3.9.6, and my platform is macOS 11.4.

Please let me know if I can provide more details to troubleshoot!

glensc:
@flammable I'm sure then it's related to cached plex requests as @twolaw suggested:
- https://github.com/Taxel/PlexTraktSync/issues/384#issuecomment-881047027

seems, it affects only new agents (I tested with legacy agent myself).

(someone) needs to debug what (extra) requests plex makes for a new agent. as for legacy agents, all movie details were given with library feed, and that request is not cached.

as a workaround use any version that's older than 0.13.0

flammable:
That sounds reasonable. I'm facing both issues, so it was hard to tell if they're part of the same problem: when I watch new movies and TV shows, they don't sync to Trakt unless I delete the cache (this issue), and every time I perform a sync, some TV shows are repeatedly marked as watched (#384).

Are there any logs I can provide to help solve this issue? Is there a way to enable debugging in the script to provide more detailed output? Thanks for all of your work on this.

glensc:
I think it's best to mail down a commit that causes the breakage. you can try to use --show or --movie parameter to sync only a specific item. but beware that those options ignore configured libraries, and perform search and sync on all libraries (ignoring config.json)

[...]

Issue #384, opened 2021-07-15
Certain movies "watched status" are being added multiple times to Plex history


rfgamaral:
### The problem

Certain movies "watched status" are being added multiple times, here's an example:

```
amaral@HYPERCUBE:~/.local/scripts/PlexTraktSync$ pipenv run python main.py sync --sync movies
Loading .env environment variables...
INFO: PlexTraktSync [5dfc1801: Merge pull request #382 from glensc/Pipfile.lock-regenerate @2021-07-14 11:58:37]
INFO: Syncing with Plex redacted@domain.com and Trakt <redacted>
Sync Movies: True
Sync Shows: False
INFO: Loaded Trakt lists in 3.8 seconds
Plex Server version: 1.23.5.4801-bc4dd58bb, updated at: 2021-07-15 12:30:49
Server has 2 libraries: ['Movies', 'TV Shows']
Processing Movies  [###---------------------------------]  22/229  00:00:23INFO: Marking as watched in Plex: <tmdb:497698:<Movie:5121:Black-Widow>>
Processing Movies  [###############---------------------]  100/229  00:00:14INFO: Marking as watched in Plex: <tmdb:512200:<Movie:3118:Jumanji:-The-Next-Le>>
Processing Movies  [###################################-]  226/229  00:00:00INFO: Marking as watched in Plex: <tmdb:637649:<Movie:3249:Wrath-of-Man>>
Processing Movies  [####################################]  229/229
INFO: Movies processed in 30.2 seconds
INFO: Updated plex watchlist in 0.0 seconds
INFO: Completed full sync in 34.1 seconds
amaral@HYPERCUBE:~/.local/scripts/PlexTraktSync$ pipenv run python main.py sync --sync movies
Loading .env environment variables...
INFO: PlexTraktSync [5dfc1801: Merge pull request #382 from glensc/Pipfile.lock-regenerate @2021-07-14 11:58:37]
INFO: Syncing with Plex <redacted@domain.com> and Trakt <redacted>
Sync Movies: True
Sync Shows: False
INFO: Loaded Trakt lists in 3.0 seconds
Plex Server version: 1.23.5.4801-bc4dd58bb, updated at: 2021-07-15 12:30:49
Server has 2 libraries: ['Movies', 'TV Shows']
Processing Movies  [###---------------------------------]  22/229  00:00:23INFO: Marking as watched in Plex: <tmdb:497698:<Movie:5121:Black-Widow>>
[...]

PR #401, opened 2021-07-27
Fix caching issue during sync

mcdallas:
Fixes #384

PR #410, opened 2021-08-01
Disable cache for guids resolve

glensc:
- Fixes https://github.com/Taxel/PlexTraktSync/issues/387
- Fixes https://github.com/Taxel/PlexTraktSync/issues/384
- Replaces https://github.com/Taxel/PlexTraktSync/pull/401

Problem is that the xml feed from section.all() does not contain data for media item .guids, so data for that is called when item.guids is first accessed, lazy loaded on demand.

Ideally this should be solved in PMS or python-plexapi:
- https://github.com/pkkid/python-plexapi/issues/808

### DIFF
--- a/plex_trakt_sync/plex_api.py
+++ b/plex_trakt_sync/plex_api.py
@@ -130,6 +130,7 @@ class PlexLibraryItem:
     def guid(self):
         return self.guids[0]
 
+    @nocache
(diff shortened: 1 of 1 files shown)


mcdallas:
Looks like this fixes #384 from my testing it 👍

PR #415, opened 2021-08-03
Bring back guid error log verbosity

twolaw:
1ceb69d33b2a6f84465767357047663ce6b949a5 removed information (media type and media name) in two log messages, making more difficult for user to identify concerned media.

Exemple :
`Skipping <local://4392:<local:4392:<Movie:4392:Thomas-Vdb-Bon-Chien>>>: Provider local has no external Id`
became :
`Skipping local://4392: Provider local has no external Id`

This PR reverts it to bring back the full information in log messages.

### DIFF
--- a/plex_trakt_sync/media.py
+++ b/plex_trakt_sync/media.py
@@ -130,13 +130,13 @@ class MediaFactory:
 
     def resolve_guid(self, guid: PlexGuid, tm=None):
         if guid.provider in ["local", "none", "agents.none"]:
-            logger.warning(f"Skipping {guid}: Provider {guid.provider} has no external Id")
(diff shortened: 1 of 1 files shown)


glensc:
btw, you can use `inspect 4392` in those cases. gives a lot of info.

glensc:
this change is not accurate. the `guid` in error could be any entry from `pm.guids`, but formatting of `str(pm)` always takes first `pm.guid`: `guids[0]`.

a more accurate fix would be to revert this PR and add `__str__` method to `PlexGuid` that returns `guid.provider`, `guid.id`, `guid.pm.item.title` (or the like).

PR #424, opened 2021-08-06
Test against resolution width on collection metadata

chrillep:
test_collection_metadata.py fails as testdata has incorrect object

Fixes resolution in testdata


glensc:
Describe the problem in PR body please.

PR #426, opened 2021-08-08
Set timezone in collection metadata test

chrillep:
collection metadata test fails because of local timezone in datetime object

this sets utc timezone when creating datetime object


glensc:
much apprechiate!

chrillep:
this brings me closer to automating tests in a pipeline =) 
PR https://github.com/chrillep/PlexTraktSync/pull/24

Issue #431, opened 2021-08-09
increasing memory usage - potentially leak?


kimpenhaus:
### The problem

While I am facing an OAuth problem on my main machine (see: #392) I switched to the [docker image](https://ghcr.io/taxel/plextraktsync:latest) which started to be published couple of days ago. I have two libraries to be synchronized (movies, tv-shows) of mid size (~ 4.5k movies and ~ 324 tv-shows). Normally when building up new docker images I restrict them to ensure the docker host isn't affected. So I gave that docker 256MB memory - as I think that synchronizing from plex to trakt shouldn't be a memory consuming task. I then faced several restarts because memory went out. To have a complete run I needed to give that docker  2GB memory.

The memory chart for a complete run looks like:

<img width="579" alt="image" src="https://user-images.githubusercontent.com/575109/128664563-14001699-af0f-49e0-8c2d-cfcf842b4bed.png">

First I thought this might because of building up the cache but another run with complete cache looks quite similar:

<img width="575" alt="image" src="https://user-images.githubusercontent.com/575109/128664624-37804a1d-325c-4f73-a41e-17c5985d22b0.png">

You can see the drop after the first run and then increasing memory usage again.

The docker image used - is tagged latest and was of version 0.13.12 during the run.

### Expected behavior

I am not sure if this is by design but I would assume a synchronizing task to consume less and more constant memory. Main problem I see is with increasing library size memory consumption would also increase during synchronization, so that a docker image can't be restricted (otherwise would start to fail at some point)

### Steps to reproduce the behavior

I would think just start synchronizing a library and see memory consumption and which objects are not garbage collected after usage... 

### Workarounds

- increase memory

### Environment

- Version: 0.13.12
- Python version: 3.9.6
- Platform: docker

mcdallas:
It’s probably the unbound lru_cache on the @memoize decorator. This project is fixated on sync speed and pretty much every single function call is cached either in memory or on disk

glensc:
you can clear the cache, like this:
- https://github.com/Taxel/PlexTraktSync/blob/0.13.14/plex_trakt_sync/plex_api.py#L414

but it's tricky to figure when to call and how on what objects.


also, it's worth reviewing not to call memorize on things only called once:
- https://github.com/Taxel/PlexTraktSync/blob/0.13.14/plex_trakt_sync/plex_api.py#L251

glensc:
Some blog post about memory usage measurements:
[...]

PR #436, opened 2021-08-11
Buffer results in section.all() by page size

glensc:
Instead of getting all results via [plexapi.library.LibrarySection.all()], implement pager locally to buffer at most `X_PLEX_CONTAINER_SIZE` items at a time.

This should reduce memory consumption to a constant value.

Users with a large library may find the progress bar being more responsive: it starts to process items sooner.

[plexapi.library.LibrarySection.all()]: https://python-plexapi.readthedocs.io/en/latest/modules/library.html#plexapi.library.LibrarySection.all

NOTE: This uses private method section._buildSearchKey, which may break with any PlexAPI update. Instead, such yielding by page size should be implemented in PlexAPI itself.

About the implementation: it's important to have `fetch_items` with `@nocache` as a separate method (instead in method `all()` directly), as using `yield` in `all()` would pop out the no-cache behavior.

Refs:
- https://github.com/Taxel/PlexTraktSync/issues/431

### DIFF
--- a/plex_trakt_sync/plex_api.py
+++ b/plex_trakt_sync/plex_api.py
@@ -385,8 +385,8 @@ class PlexLibrarySection:
         return self.section.fetchItems(key, container_start=start, container_size=size)
 
     def items(self, max_items: int):
-        for item in (PlexLibraryItem(x) for x in self.all(max_items)):
-            yield item
+        for item in self.all(max_items):
+            yield PlexLibraryItem(item)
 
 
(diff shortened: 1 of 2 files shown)


glensc:
While timing is not important in this PR, I was curious any way so I measured

- before: 7.1s, 6.7s, 6.8s, 6.8s, 6.6s
- after: 6.9s, 6.3s, 6.6s, 6.2s, 6.5s

PR #439, opened 2021-08-12
Add `no-progress-bar` option to the sync and unmatched commands

Matty666:
to allow them to not use the progress bar, this can be used to simplify the output for non-interactive situations

### DIFF
--- a/plex_trakt_sync/commands/unmatched.py
+++ b/plex_trakt_sync/commands/unmatched.py
@@ -6,8 +6,15 @@ from plex_trakt_sync.factory import factory
 from plex_trakt_sync.walker import Walker
 
 
+@click.option(
+    "--no-progress-bar", "no_progress_bar",
+    type=bool,
+    default=False,
+    is_flag=True,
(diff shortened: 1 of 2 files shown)


glensc:
how about `--no-progess` or `--no-progessbar` instead?

Matty666:
Sure! I'll change it to that!

plex_trakt_sync/commands/unmatched.py
```diff
@@ -5,17 +5,24 @@
 from plex_trakt_sync.factory import factory
 from plex_trakt_sync.walker import Walker
 
-
```
glensc, line 8: restore two newlines here. some PEP style.

glensc:
update PR title to be accurate, and minor style fix. then this is ready to go!

plex_trakt_sync/commands/unmatched.py
Matty666: yeah sorry i missed that! done

PR #444, opened 2021-08-16
Drop TqdmLoggingHandler class

glensc:
`tqdm.tqdm` implements necessary `.write()` method, so it can be used as a stream.

refs https://github.com/Taxel/PlexTraktSync/pull/386

cc @mcdallas


mcdallas (approved):
👍

PR #457, opened 2021-08-29
Add env variables to override path for logs/configs/cache

glensc:
This is to simplify docker based deployments putting all files that are updated to a single directory.


glensc:
https://github.com/Taxel/PlexTraktSync/pull/458 uses /app/config for those configs, logs, cache paths. should we change app default as well?

PR #464, opened 2021-09-03
Print Plex play history on inspect

glensc:


plex_trakt_sync/plex_api.py
```diff
@@ -491,6 +501,21 @@ def delete_playlist(self, name: str):
         except (NotFound, BadRequest):
             logger.debug(f"Playlist '{name}' not found, so it could not be deleted")
 
+    @nocache
```
glensc, line 504: `@nocache` and yield don't play well:
- https://github.com/Taxel/PlexTraktSync/pull/466

PR #470, opened 2021-09-04
Change Plex fallback url if run in docker

twolaw:
`PLEX_FALLBACKURL=http://localhost:32400` in **.env** is irrelevant if script runs inside a docker container. In this case, _localhost_ refers to the container only but Plex is not in the same container.

This PR adds code to detect (during Plex credentials setup) if run inside docker, find container host ip adress and write `PLEX_FALLBACKURL=http://ip_address:32400` instead. 
172.17.0.1 is the default ip address of container's host.

This is not an important issue because the PLEX_FALLBACKURL is used only as last resort, when Plex connection fails.

~~Docker detection is made through _/.dockerenv_ file or _/proc/self/cgroup_ content.~~
~~_Another solution would be to create a dedicated environment variable (such as RUN_IN_DOCKER) at container creation so it could be read during setup._~~

### DIFF
--- a/plex_trakt_sync/commands/plex_login.py
+++ b/plex_trakt_sync/commands/plex_login.py
@@ -2,6 +2,8 @@ from datetime import datetime, timedelta
 from functools import partial
 from os import environ
 from typing import List
+from subprocess import check_output
+import socket
 
 import click
 from click import Choice
@@ -171,7 +173,17 @@ def plex_login(username, password):
     CONFIG["PLEX_USERNAME"] = user
     CONFIG["PLEX_TOKEN"] = token
     CONFIG["PLEX_BASEURL"] = plex._baseurl
-    CONFIG["PLEX_FALLBACKURL"] = "http://localhost:32400"
+    if environ.get("PTS_IN_DOCKER"):
+        try:
+            host_ip = socket.gethostbyname("host.docker.internal")
+        except socket.gaierror:
+            try:
+                host_ip = check_output("ip -4 route show default | awk '{ print $3 }'", shell=True).decode().rstrip()
+            except:
+                host_ip = "172.17.0.1"
+        CONFIG["PLEX_FALLBACKURL"] = f"http://{host_ip}:32400"
+    else:
+        CONFIG["PLEX_FALLBACKURL"] = "http://localhost:32400"
     CONFIG.save()
 
     click.echo(SUCCESS_MESSAGE)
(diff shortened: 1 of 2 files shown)


glensc:
I think it's better just add ENV to Dockerfile: `PTS_IN_DOCKER=1`

glensc:
also, the default is not that universal, so maybe there's little point in all this?

```
➔ docker-compose run --entrypoint 'sh -c'  plex-trakt-sync  'ip addr show'|grep inet
Creating plextraktsync_plex-trakt-sync_run ... done
    inet 127.0.0.1/8 scope host lo
    inet 172.21.0.2/16 brd 172.21.255.255 scope global eth0
```

twolaw:
Updated with ENV in Dockerfile and host ip read at setup.

plex_trakt_sync/commands/plex_login.py
```diff
@@ -171,7 +171,14 @@ def plex_login(username, password):
     CONFIG["PLEX_USERNAME"] = user
     CONFIG["PLEX_TOKEN"] = token
     CONFIG["PLEX_BASEURL"] = plex._baseurl
-    CONFIG["PLEX_FALLBACKURL"] = "http://localhost:32400"
+    if environ.get("PTS_IN_DOCKER"):
+        try:
+            host_ip = check_output("ip -4 route show default | awk '{ print $3 }'", shell=True).decode().rstrip()
```
glensc, line 176: perhaps we should try host.docker.internal name first. it exists in docker for mac and docker for windows

- https://docs.docker.com/desktop/mac/networking/

```
➔ docker-compose run --entrypoint 'sh -c'  plex-trakt-sync  'getent hosts host.docker.internal'
Creating plextraktsync_plex-trakt-sync_run ... done
192.168.65.2      host.docker.internal  host.docker.internal
```

```
➔ docker-compose run --entrypoint 'python -c'  plex-trakt-sync  'import socket; print(socket.gethostbyname("host.docker.internal"));'
Creating plextraktsync_plex-trakt-sync_run ... done
192.168.65.2
```

glensc:
LGTM

glensc (approved):

twolaw:
Do you think I need to force-push again because I used 4 spaces instead of TAB in Dockerfile ENV command ?

glensc:
@twolaw do if you wish. I gave my LGTM for merge already.

Issue #411, opened 2021-08-01
Feature request: official docker repository


xonvanetta:
### Feature description

Having an official docker repository with the build tags ready to be used.   

### Use case

Instead of having to install python and all the requirements we could just use it in a container and run it without needing to install it. A benefit to this is that if someone is saying they are using docker with tag X it easier to troubleshoot than to guess what python version they are using and if the installation was correct. 

### Workarounds

https://hub.docker.com/r/kahooli/plextraktsync
Not up to date, also not using tags which i am no fan of.  

### Plan to implement

I could try but i am not fluent in python with what is best practices for running code in containers.
There was a Dockerfile if its ready to use would only need to configure github to push finished image to a repository, and some docs so it would be easy to understand

Scales82:
yes I would 100% love to have this. a docker compose with it would be great, my containers would automatically update also

glensc:
The Dockerfile already exists here, someone needs to provide a docker compose and GitHub action to build and push docker images.

glensc:
ps: python version and incorrect installation is rarely the source of problems :)

chrillep:
only issue would be owner of the docker hub repo 😊.


https://docs.github.com/en/actions/guides/publishing-docker-images#publishing-images-to-docker-hub-and-github-packages

rfgamaral:
> 
> 
> only issue would be owner of the docker hub repo 😊.
> 
> [docs.github.com/en/actions/guides/publishing-docker-images#publishing-images-to-docker-hub-and-github-packages](https://docs.github.com/en/actions/guides/publishing-docker-images#publishing-images-to-docker-hub-and-github-packages)

It should be published to ghcr.io instead.

KaHooli:
While I've kept my duplicate GitHub repository current, the auto build feature on Docker Hub is no longer free, so it isn't keeping my repo up to date.

I'm looking into GitHub Actions for publishing to ghcr for my own education, but I agree it would be better if done on the official GitHub repo.

glensc:
Well, this issue is exactly for that. submit your PRs!

- https://github.com/Taxel/PlexTraktSync/issues/411#issuecomment-890745171

glensc:
Docker images are published now:
- https://github.com/Taxel/PlexTraktSync/pull/413
- https://github.com/Taxel/PlexTraktSync/pull/423

This should be documented now.

also, docker compose config should be added and such invocation be preferred.

iFredOS:
[...]

PR #458, opened 2021-08-29
Add docker compose config

glensc:
Usage:

```
docker-compose run --rm plex-trakt-sync
```

Requires https://github.com/Taxel/PlexTraktSync/pull/457 to be merged first

refs:
- https://github.com/Taxel/PlexTraktSync/issues/411

PR #471, opened 2021-09-04
Docker Compose instructions

KaHooli:
Attempt requested under issue #411

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,11 +148,11 @@ This will show you how to set up PlexTraktSync for one user on a Plex Media Serv
 First we need to create the configuration files needed.  This can be done by running the following command line...
 ```bash
 docker run \
-	-it \ # This runs an interactive session
-	--rm \ # This 
-	--name plextraktsync-james \
-	--network t2_proxy \
-	-v /home/james/docker/appdata/plextraktsync-james2/config:/app/config \
+	-it \ # Runs session as interactive
+	--rm \ # Removes container after closing
+	--name plextraktsync-jimbob \ # Name of the session, helps with diagnosing issues if they happen
+	--network t2_proxy \ # Name of the same network that your Plex 
+	-v /home/james/docker/appdata/plextraktsync-jimbob/config:/app/config \
 	-e PTS_CONFIG_DIR=/app/config \
 	-e PTS_CACHE_DIR=/app/config \
 	-e PTS_LOG_DIR=/app/config \
@@ -174,6 +174,6 @@ docker run \
  - The PyTrakt API keys are not stored securely, so if you do not want to have
    a file containing those on your harddrive, you can not use this project.
 <!--stackedit_data:
-eyJoaXN0b3J5IjpbMTIyMDYyNjkzMCwxODkzOTIwNzMyLDI5OT
+eyJoaXN0b3J5IjpbLTY1NjMwODE1MywxODkzOTIwNzMyLDI5OT
 M0ODM0M119
 -->
--- a/README.md
+++ b/README.md
@@ -143,7 +143,9 @@ by user is not implemented [#324].
 ## Docker Compose
 
 This will show you how to set up PlexTraktSync for one user on a Plex Media Server for Docker Compose.  Duplicate this for each additional user wanting to sync to Trakt.
+
 ### Setup
+
 First we need to create the configuration files needed.  This can be done by running the following command line...
 ```bash
 docker run \
@@ -170,8 +172,9 @@ Keep running this until you've completed your first successful sync between Plex
  1. Setup link with Plex Media Server
  2. Setup link with Trakt
  3. Sync between Plex Media Server and Trakt.
- 
+
 ### Docker Compose
+
 Once you have your configuration files created, you need to make sure they have the correct permissions to be used by docker compose as they've probably been created with ```root:root```.
 
 Docker normally uses the local user and the docker group, so run ```sudo chown user:docker /home/user/docker/appdata/plextraktsync-john -R``` .  Check your own setup incase it is different.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -140,6 +140,64 @@ by user is not implemented [#324].
...
+    depends_on:
+      - plex
+    restart: unless-stopped
+    volumes:
+      - $DOCKERDIR/appdata/plextraktsync-john/config:/app/config
+    environment:
+      - PTS_CONFIG_DIR=/app/config
+      - PTS_CACHE_DIR=/app/config
+      - PTS_LOG_DIR=/app/config
+    entrypoint: ["python3", "-m", "plex_trakt_sync", "watch"]
```
twolaw, line 198: I suggest to show a `sync` command as example instead of `watch`.
`sync` is the standard command for full plex->trakt and trakt->plex status update.
`watch` is a specific command.

README.md
```diff
@@ -140,6 +140,64 @@ by user is not implemented [#324].
...
+    image: ghcr.io/taxel/plextraktsync:latest
+    container_name: plextraktsync-john
+    networks:
+      - t2_proxy
+    depends_on:
+      - plex
+    restart: unless-stopped
+    volumes:
+      - $DOCKERDIR/appdata/plextraktsync-john/config:/app/config
+    environment:
```
twolaw, line 194: I used to add a timezone environment variable to have local time on logs instead of UTC time.
`- TZ=Europe/London`
Users get it [here](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).

twolaw:
Nice job.
Can't all those commits be only one commit ?

KaHooli:
Thanks.

I used an online WYSIWYG editor for the markdown language, as I was pretty tired & not in a clear headspace to work it all out quickly.  It unfortunately committed after every little change.

End result is something I am fairly happy with.

README.md
KaHooli: Isn't ```sync``` a once off, ie will close the container after it completes once?  And ```watch``` keeps watching and reporting progress and completion, keeping the container live?

twolaw: Yes, but `watch` does not perform a full sync.
I think the example given in readme should be standard as much as possible.
I suggest to remove `watch` from entrypoint (just like [Dockerfile](https://github.com/Taxel/PlexTraktSync/blob/main/Dockerfile) is, no command means sync).
[Readme](https://github.com/Taxel/PlexTraktSync#setup) explains the use of a cronjob to repeat the sync.

**Keeping container alive :** @Taxel [talked](https://github.com/Taxel/PlexTraktSync/discussions/88) about a cronjob inside the container to keep it alive and make it run `sync` every x hours. It's a good idea, maybe for a next PR.

KaHooli: A full sync is performed on the third run of the container in the Setup phase.
I then run the container in watch mode from Docker Compose as that suits my neads.
Feel free to modify what I've written.

README.md
```diff
@@ -140,6 +140,64 @@ by user is not implemented [#324].
 
 [#324]: https://github.com/Taxel/PlexTraktSync/issues/324
 
+## Docker Compose
```
glensc, line 143: duplicate docker compose section. and this section isn't even about docker compose.

README.md
glensc, line 198: dob't override entry-point. use `command: watch` instead

glensc:
please reorder documentation that you show docker-compose example first, as it's simple and straightforward, like described in this PR:
- https://github.com/Taxel/PlexTraktSync/pull/458

glensc:
I read the changes again, and the `docker run` can be replaced with `docker-compose run`:
- https://github.com/Taxel/PlexTraktSync/pull/458

so, restructure with:
1. creating docker-compose config
2. telling to run it multiple times
3. telling about sync command (default) and watch command

glensc:
also, the chown is not necessary, currently container runs as root. so remove the chown text to keep things simple.

KaHooli:
> please reorder documentation that you show docker-compose example first, as it's simple and straightforward, like described in this PR:
> 
> - https://github.com/Taxel/PlexTraktSync/pull/458
> 
> 

The problem is that you need the configuration files created before you run it under docker compose.

KaHooli:
> also, the chown is not necessary, currently container runs as root. so remove the chown text to keep things simple.

It wouldn't work for me unless I did this step, I got errors about not being able to access files.

glensc:
> > also, the chown is not necessary, currently container runs as root. so remove the chown text to keep things simple.
> 
> It wouldn't work for me unless I did this step, I got errors about not being able to access files.

please show those errors and details of your environment.

glensc:
> > please reorder documentation that you show docker-compose example first, as it's simple and straightforward, like described in this PR:
> > 
> > * [Add docker compose config #458](https://github.com/Taxel/PlexTraktSync/pull/458)
> 
> The problem is that you need the configuration files created before you run it under docker compose.

what config files? share the problem detail.

README.md
glensc: @twolaw perhaps you make PR to add tzdata to dockerfile and example docker-compose

README.md
```diff
@@ -140,6 +140,64 @@ by user is not implemented [#324].
...
+    container_name: plextraktsync-john
+    networks:
+      - t2_proxy
+    depends_on:
+      - plex
+    restart: unless-stopped
+    volumes:
+      - $DOCKERDIR/appdata/plextraktsync-john/config:/app/config
+    environment:
+      - PTS_CONFIG_DIR=/app/config
```
glensc, line 195: the PTS variables are no longer needed:
- https://github.com/Taxel/PlexTraktSync/pull/472

drop them from here.

glensc:
@KaHooli when are you able to address the review comments I left?

glensc:
@KaHooli are you able to address the review comments I left?

glensc:
Replaced with https://github.com/Taxel/PlexTraktSync/pull/504

grasuoare:
Hi all.

This is my docker compose:

  plex-trakt-sync:
    image: ghcr.io/taxel/plextraktsync
    container_name: plex-trakt-sync
    restart: unless-stopped
    networks:
      - t2_proxy
    volumes:
      - $DOCKERDIR/plex-trakt-sync/config:/app/config

Could someone tell me where to insert the other data such as username, password, etc? Are there any environment variables to set?

PR #474, opened 2021-09-06
Update Docker base image to alpine 3.15

glensc:
https://alpinelinux.org/posts/Alpine-3.14.0-released.html

This requires the docker host to be fresh enough:

> The faccessat2 syscall has been enabled in musl. This can result in issues on docker hosts with older versions of docker (<20.10.0) and libseccomp (<2.4.4), which blocks this syscall.

----

https://alpinelinux.org/posts/Alpine-3.15.0-released.html

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM python:3.10-alpine3.13 AS base
+FROM python:3.10-alpine3.15 AS base
+
 WORKDIR /app
 
 # Install app depedencies


glensc:
> The faccessat2 syscall has been enabled in musl. This can result in issues on docker hosts with older versions of docker (<20.10.0) and libseccomp (<2.4.4), which blocks this syscall.

maybe not so good idea to upgrade to latest and greatest?

eidanyosoy:
is this still the case with 3.15?

glensc:
the change is since 3.14, so yes.

glensc:
@eidanyosoy if you are willing to submit a pull request updating readme that such requirement appears. I'm ok to merge this.

otherwise, I would have to deal with tons of user bug reports if they use older docker, which I would not want to do.

glensc:
let's go!

PR #477, opened 2021-09-07
Fix duplicate watched status in trakt (due to movie in multiple Plex libraries)

### LINKED ISSUE #476: Duplicate when multiple library with same movie/show
### The problem

_A description of what the bug is, including a complete traceback (if applicable)_

When a movie is in two or more libraries, for example "Movies" and "4K Movies" (which I think is quite common for people to have), it is synced twice in total, once for each library, even though the movie had only been watched once.

### Expected behavior

Only sync once if same movie/show is present in two or more libraries.

Sync a show or movie that is present in two or more libraries.

### Workarounds

Exclude all but one library of the same type, but this would mean if a movie is only in a library that is excluded, it won't be synced.

### Environment

- Version: 0.13.24
- Python version: 3.9
- Platform: Ubuntu

twolaw:
The idea of this PR :
The trakt `watched_movies` list used during sync is created when syncing the first movie of the first library.
It is later read to know the watched status of other movies, but it is memoized so not fetched from trakt server. Therefore, it can give a wrong not-watched status when a movie is scanned for the second time in the same sync session.

The idea is to update this list each time a movie is marked as watched in trakt.
With this list up-to-date, the script will know the correct watched status of a movie even if it has already been scanned in the same session.

https://github.com/Taxel/PlexTraktSync/blob/83ded352c7611ccb0399a1408c9c13b0b9581833/plex_trakt_sync/trakt_api.py#L90-L98

fixes #476

### DIFF
--- a/plex_trakt_sync/trakt_api.py
+++ b/plex_trakt_sync/trakt_api.py
@@ -181,6 +181,7 @@ class TraktApi:
     @time_limit()
     def mark_watched(self, m, time):
         m.mark_as_seen(time)
+        self.watched_movies.add(m.trakt)
 
     def add_to_collection(self, m, pm: PlexLibraryItem):
         if m.media_type == "movies":


twolaw:
I'm not sure if I really should remove the memoize decorator.

glensc:
> I'm not sure if I really should remove the memoize decorator.

read my response:
- https://github.com/Taxel/PlexTraktSync/pull/477#commitcomment-56065707

without it operates on a copy:

```py
        print(type(trakt.watched_movies))
        print(3 in trakt.watched_movies)
        print(3000 in trakt.watched_movies)
        print(len(trakt.watched_movies))
        trakt.watched_movies.add(3)
        trakt.watched_movies.add(3000)
        print(3 in trakt.watched_movies)
        print(3000 in trakt.watched_movies)
        print(len(trakt.watched_movies))
```

```
<class 'set'>
True
False
859
True
False
859
```

compare with and without the decorator yourself.

glensc:
and please, try not to force push during the review. use git fixup commits instead. use issue status Draft to prevent accidental merges.

first it's difficult to see the new changes you make, if you overwrite everything each time, and github has troubles as well, as all line comments get hidden somewhere.

I recommend using git fixup (a simple wrapper for `git commit --fixup`):
- https://github.com/keis/git-fixup

my typical flow is like this:
```
git add -p
git fixup HEAD # or git commit --fixup HEAD
git push
```

and when done with the review (or just too messy and want to clean up):

```
git fetch
git rebase -i origin/main
git push --force-with-lease
```

twolaw:
Ok thanks, i'll have to learn this.

glensc:
4b01dcb3bb29d1fe19f6a1055db356a1f89904cf looks accurate now. lets wait for user feedback.

glensc:
Some of the discussion is on the commit:
- https://github.com/Taxel/PlexTraktSync/commit/694bd79f75bc238ceb8c1edf0afbc5a8a936be59#commitcomment-56065708

that discussion is not reachable from pull request due force pushes. at least i can't find it.

glensc (approved):

PR #480, opened 2021-09-09
implementation of username filter for watch

### LINKED ISSUE #324: Add scrobble user filter function
Seems like for now, the `watch` function will scrobble all users status to my trakt account.
It would be great if we could filter users at the time of scrobble.

ImjuzCY:
fixes #324 

I'm not sure if the code is up to standard, but I tested it and it works for me.

### DIFF
--- a/plex_trakt_sync/commands/watch.py
+++ b/plex_trakt_sync/commands/watch.py
@@ -21,6 +21,8 @@ class WatchStateUpdater:
         self.plex = plex
         self.trakt = trakt
         self.scrobblers = ScrobblerCollection(trakt)
+        self.sessions_from_username = []
+        self.sessions_cleared = False
 
     def __call__(self, message):
         for pm, tm, item in self.filter_media(message):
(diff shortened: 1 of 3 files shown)


glensc:
@ImjuzCY first thing first: don't create pull request from your main (master) branch. any commits added to it will became part of the pull request. this is note for the future contributions

glensc:
@ImjuzCY this needs to be configurable.

I propose adding a new key to "config.default.json", which defaults to `null` meaning current behavior, and if it's array, then it's a list of usernames whom to watch scribble watches.

ImjuzCY:
@glensc should I close this pull request first and implement them on another branch first then open another pull request?

glensc:
@ImjuzCY is up to you, you can just add more commits to your "main" branch, they will end up in this pull request.

twolaw:
In my opinion, such option is not needed.
I think the only expected behaviour of `watch` command is to scrobble the Plex account of the user. No option. Simple.

twolaw:
> a list of usernames whom to watch scribble watches

What would be the use case of this ?

glensc:
> > a list of usernames whom to watch scrobble watches
> 
> What would be the use case of this ?

I share a TV, so I want both account plays to be accounted to the same trakt account. I'm fine for having no filter at all, but thought maybe not limit to single account when making it customizable.

plex_trakt_sync/commands/watch.py
```diff
@@ -30,8 +32,35 @@ def __call__(self, message):
...
+            
+            
+            if factory.config()['watch']['username_filter']:
+                plex_sessions = self.plex.get_sessions()
+                current_sessionKeys = []
+                
+                for session in plex_sessions:
+                    current_sessionKeys.append(session.sessionKey)
+                
+                if 1 in current_sessionKeys and not self.sessions_cleared:
```
glensc, line 44: please explain what is this magic `1`

ImjuzCY: If there is a playing session when the Plex server restarts, it's sessionKey resets to 1 when the server comes back up. I can't find a way to be notified when the server restarts, so I figure if there is a sessionKey == 1, it means that the server has restarted so the sessions_from_username list has to be cleared and updated with the new sessionKey.

glensc: please post those responses, as asked already:
- https://github.com/Taxel/PlexTraktSync/pull/480#issuecomment-916071825

ImjuzCY: I do not understand what you mean by responses. Do you mean the console output?

```
Found <tmdb:170258:<Episode:326476:The-Office-s00e09>>
<Episode:326476:The-Office-s00e09>: 0.000000% Watched: False, LastViewed: None
New session from imjuzcy: 36
From username's session 36
State: playing
Found <tmdb:170258:<Episode:326476:The-Office-s00e09>>
<Episode:326476:The-Office-s00e09>: 2.121141% Watched: False, LastViewed: None
From username's session 36
State: paused
Found <tmdb:170258:<Episode:326476:The-Office-s00e09>>
<Episode:326476:The-Office-s00e09>: 2.356823% Watched: False, LastViewed: None
From username's session 36
State: stopped
Found <tmdb:170258:<Episode:326476:The-Office-s00e09>>
<Episode:326476:The-Office-s00e09>: 2.356823% Watched: False, LastViewed: None
From username's session 36
Remaining sessions from imjuzcy: []
```

Do you think it's too much output?

glensc: I mean those api responses. `print(self.plex.get_sessions())`

ImjuzCY: `print(self.plex.get_sessions())` :
```
[<Episode:192182:The-Office-s02e04>]
```

This `Episode` object is different than the `movie` object in the function.
`print(self.plex.get_sessions()[0].usernames)` : 
```
['imjuzcy']
```

`print(movie.usernames)` :
```
[]
```

ImjuzCY:
@glensc I figured out a way to avoid polling the Plex server every second to get sessions. Now it just have to poll the server once when a new session starts. Please review this

glensc:
Re-created the same logic on the current codebase:
- https://github.com/Taxel/PlexTraktSync/pull/480

please test that it behaves similarly.

glensc:
Since https://github.com/Taxel/PlexTraktSync/pull/502 is merged, you may want to submit documentation for the feature (change to README.md)

Issue #448, opened 2021-08-20
Feature request: Use STDERR as its intended


oxivanisher:
The current implementation (progress bar on stderr) makes the usage of a cronjob problematic.

Currently, you have to redirect stderr to stdout to net get a email for every cron run. Also this way, you don't get notified if something was going wrong. It would IMHO be much better, if there is a verbose switch to control how much output a normal run creates and only errors are sent to stderr.

glensc:
Use `--no-progress-bar`:
- https://github.com/Taxel/PlexTraktSync/pull/439

You're welcome to send PR documenting it.

twolaw:
The question remains unanswered.

Shouldn't tqdm progressbar be redirected to stdout ?
https://github.com/Taxel/PlexTraktSync/blob/6fce2adf2d8888711f38395060ed131991dac1c6/plex_trakt_sync/walker.py#L122

```python
pb = self._progressbar(iterable, file=sys.stdout, **kwargs)
```

Ref:
- https://tqdm.github.io/docs/tqdm/

PR #481, opened 2021-09-10
use unbuffered output for docker

ImjuzCY:
So that output can be shown in `docker logs`


glensc:
Since docker docs use env, let's go with that for now:
- https://docs.docker.com/samples/django/

```dockerfile
ENV PYTHONUNBUFFERED=1
```

twolaw:
Speaking about output :
- https://github.com/Taxel/PlexTraktSync/issues/448#issuecomment-903120101

PR #482, opened 2021-09-11
Fix duplicate watched status in trakt (due to tv episode in multiple Plex libraries)

### LINKED ISSUE #476: Duplicate when multiple library with same movie/show

twolaw:
Same as #477 but for tv shows.

fixes #476 

note : [pytrakt_extensions.py](https://github.com/Taxel/PlexTraktSync/blob/main/plex_trakt_sync/pytrakt_extensions.py) uses many `if value in dict.keys():` which [seems](https://stackabuse.com/python-check-if-key-exists-in-dictionary/) to be slower than `if value in dict:`
note 2 : please @glensc do not merge in a hurry.

### DIFF
--- a/plex_trakt_sync/pytrakt_extensions.py
+++ b/plex_trakt_sync/pytrakt_extensions.py
@@ -140,6 +140,17 @@ class AllWatchedShows():
             return False
         return self.shows[trakt_id].seasons[season].get_completed(episode)
 
+    def add(self, trakt_id, season, episode):
+        episode_prog = {"number":episode, "completed":True}
+        season_prog = {"number":season, "episodes":[episode_prog]}
+        if trakt_id in self.shows:
+            if season in self.shows[trakt_id].seasons:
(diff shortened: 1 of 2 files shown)


glensc:
> note 2 : please @glensc do not merge in a hurry.

use GitHub draft state to indicate PR is not ready for merge.

PR #490, opened 2021-09-17
Reorganised readme

sirloinofbeef:
Reorganised readme so it is easier to follow.
- Each step is now a dot point, with alternative options (such as Pipenv) being listed with `or` to make it clear.
- New `Commands` section gathers together Sync, Unmatched & Watch into one logical space, to which further commands can be added as needed.
- Previous `Notes` footnotes have been redistributed to their logical sections for easier reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ This project adds a two-way-sync between trakt.tv and Plex Media Server. It
 requires a trakt.tv account but no Plex premium and no Trakt VIP subscriptions,
 unlike the Plex app provided by Trakt.
 
+Note: The PyTrakt API keys are not stored securely, so if you do not want to have a file containing those on your harddrive, you can not use this project.
+
 **To contribute, please find issues with the [`help-wanted`](https://github.com/Taxel/PlexTraktSync/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22) label, thank you.**
 
 [python-versions-badge]: https://img.shields.io/badge/python-3.7%20%7C%203.8%20%7C%203.9-blue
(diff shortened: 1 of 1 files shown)


glensc:
LGTM, @twolaw what you?

twolaw:
In `Setup`, the creation of Trakt App must remain the first step. And then user runs the script.

sirloinofbeef:
> In `Setup`, the creation of Trakt App must remain the first step. And then user runs the script.

Happy to move it if people feel strongly about it.

The creation of a Trakt app is explained within the script though at the appropriate point using the exact same text.
Why double up on the instruction?
It makes more sense (to me) to have it alongside the other info we have there, which is also covered in-script during setup.

twolaw:
> The creation of a Trakt app is explained within the script though at the appropriate point using the exact same text.
> Why double up on the instruction?
> It makes more sense (to me) to have it alongside the other info we have there, which is also covered in-script during setup.

In the script, it is a reminder just in case the user forgot : `If you do not have...` . The original order is to create the App before launching the script.
Asking user to do both in parallel (run script, then stop what you're doing on it to go create an app on the web) is more complicated IMHO.

Also, like you I'm not found of [doubling](https://github.com/Taxel/PlexTraktSync/commit/9a0efc0c4b6fa561b92429e06ee8f01049e8b9ca) the instructions. I think readme file is the only correct place for it.
I think script reminder should only say `If you do not have a Trakt client ID and secret, please read instructions in readme file.`

sirloinofbeef:
> Asking user to do both in parallel (run script, then stop what you're doing on it to go create an app on the web) is more complicated IMHO.
A compelling point well made. I'll make the change.

> I think script reminder should only say `If you do not have a Trakt client ID and secret, please read instructions in readme file.`
My preference is to keep it with the instructions, as if people have missed this step they don't have to go hunting for the info.

twolaw:
Ok, it looks good to me 👍

PR #493, opened 2021-09-21
Use LazyUserList to fetch Trakt Lists faster

### LINKED ISSUE #489: Fails to retrive Trakt List, if it has Episodes Instead of Shows
### The problem

During the step where Trakt Lists are downloaded, One list keeps failing, and I Belive it is because it has episodes instead of shows. This is the list in particular
https://trakt.tv/users/grendelcrusher/lists/doctor-who-universe-watching-order
`WARNING: Failed to get list Doctor Who Universe Watching Order by user grendelcrusher`

### Expected behavior

Download said list and add it to plex with available episodes

Have list likes on Trakt, and run PlexTraktSync

### Workarounds

No

### Environment

- Version: 0.13.27
- Python version: 3.9.7
- Platform: Windows

twolaw:
This PR creates **LazyUserList** class being inherited from trakt **UserList** class but without requesting each item of the list to Trakt API.
Fetching Trakt lists should be much faster and not crashing if above 1000 items (GET rate limit of trakt API).

fixes #489

### DIFF
--- a/plex_trakt_sync/trakt_list_util.py
+++ b/plex_trakt_sync/trakt_list_util.py
@@ -5,9 +5,36 @@ from trakt.errors import NotFoundException, OAuthException
 from trakt.movies import Movie
 from trakt.tv import TVEpisode
 from trakt.users import UserList
+from trakt.core import get
+from trakt.utils import slugify, extract_ids
 
 from plex_trakt_sync.logging import logger
 from plex_trakt_sync.plex_api import PlexApi
+
+
+class LazyUserList(UserList):
+    @get
+    def get_items(self):
+        data = yield 'users/{user}/lists/{id}/items'.format(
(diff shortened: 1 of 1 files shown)


plex_trakt_sync/trakt_list_util.py
```diff
@@ -5,17 +5,45 @@
...
+                continue
+            item_type = item['type']
+            item_data = item.pop(item_type)
+            extract_ids(item_data)
+            self._items.append((item_type + 's', item_data['trakt']))
+        yield self._items
+
+    @classmethod
+    @get
+    def _get(cls, title, creator):
```
glensc, line 31: `_get` override should not be needed after https://github.com/glensc/python-pytrakt/pull/3

Issue #495, opened 2021-09-24
Scrobbling not working in watch mode running inside container


isaacrlevin:
### The problem

Scrobbling not occurring when Sync is running in docker container when watch mode is turned on

### Expected behavior

Scrobbling completes in watch mode

### Steps to reproduce the behavior

**NOTE I have Plex running in a docker container as well, on a Traefik network where the network IP of Plex is *192.168.50.200***

I successfully initialized sync running in an intermediate docker container with 

```
docker run   -it --rm   --name plextraktsync   --network traefik_proxy  -v /mnt/data/config/plextraktsync/config:/app/config   ghcr.io/taxel/plextraktsync:latest
```

With my config set, I than created a container with docker-compose

```
version: "3.7"
services:
  plextraktsync:
    image: ghcr.io/taxel/plextraktsync:main
    container_name: plextraktsync
    networks:
      - traefik_proxy
    restart: unless-stopped
    volumes:
      - /mnt/data/config/plextraktsync/config:/app/config
    entrypoint: ["python3", "-m", "plex_trakt_sync", "watch"]

networks:
  traefik_proxy:
    external: true

```

Looking in the docker logs, all I see is 

```
today at 9:35:41 PM Listening for events!
today at 9:35:41 PM INFO: Starting AlertListener: ws://192.168.50.200:32400/:/websockets/notifications?X-Plex-Token=<hidden>
```

and when I mark an item as played in Plex, no sync is triggered. Looking in the Plex logs, I see entries like below is I mark and unmark as played

```
Completed: [192.168.86.59:57567] 200 GET /:/unscrobble?identifier=com.plexapp.plugins.library&key=2077 (7 live) GZIP 23ms 195 bytes (pipelined: 1)
Completed: [192.168.86.59:64610] 200 GET /:/scrobble?identifier=com.plexapp.plugins.library&key=2077 (7 live) GZIP 6ms 195 bytes (pipelined: 1)
```

Is there a way for me to investigate this more. It seems like the listener does not recognize the above requests.

### Environment

- Version: main
- Python version: 3.9
- Platform: Docker

glensc:
This looks like a feature request rather than a bug.

So you want "watch" command to recognize actions like you "mark as watched" from Plex Web?

isaacrlevin:
Maybe I don't understand what scrobbling is. Does it only listen to the we hook that fires when you press play on a video?

glensc:
Scrobbing definition is copied into readme:
[...]

PR #499, opened 2021-09-25
Feature: Add example ActivityNotification to watch command

glensc:
This shows that watching "play" / "unplayed" events from Plex Media Server is possible for https://github.com/Taxel/PlexTraktSync/issues/495

This prints:

```
Activity: <tmdb:xxx>: Watched: Plex: True, Trakt: False
Activity: <tmdb:xxx>: Watched: Plex: False, Trakt: False
```

### DIFF
--- a/plex_trakt_sync/commands/watch.py
+++ b/plex_trakt_sync/commands/watch.py
@@ -25,8 +25,10 @@ class WatchStateUpdater:
         self.mf = mf
         self.scrobblers = ScrobblerCollection(trakt)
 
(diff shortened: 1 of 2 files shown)


glensc:
Seems this activity is picked up not just on play/pause, but also on when I open move in plex web, and also when play is finished on some other tab.

so, I'll leave this as an example or building block where someone else can build implementation on.

glensc:
If someone wishes to add actual code behind this event, probably should make it configurable, as it may create unwanted trakt records.

PR #502, opened 2021-09-25
Add username filter to watch command

glensc:
To enable this feature, change `config.json`:
```json
    "watch": {
        "username_filter": true
    },
```

NOTE: it's important that you log in with plex account name, not email. i.e `PLEX_USERNAME` must be the account name.

Fixes https://github.com/Taxel/PlexTraktSync/issues/324, Replaces https://github.com/Taxel/PlexTraktSync/pull/480

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -13,6 +13,9 @@
         "collection": true,
         "ratings": true
     },
+    "watch": {
+        "username_filter": false
+    },
     "xbmc-providers": {
         "movies": "imdb",
         "shows": "tvdb"
--- a/plex_trakt_sync/commands/watch.py
+++ b/plex_trakt_sync/commands/watch.py
@@ -18,6 +18,26 @@ class ScrobblerCollection(dict):
         return value
 
 
+class SessionCollection(dict):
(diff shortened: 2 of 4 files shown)


glensc:
NOTE: plex media server restarts are not handled gracefully, but it works more or less, as likely the same id is missing, so it will re-fetch the whole list.

it definitely gets things wrong when server is restarted and user1 plays one movie, and server is restarted again and user2 plays some other movie as they both have session_id=1 cached.

I plan to investigate this further in a separate PR and add event to support server restarts

twolaw:
Do you plan to update the `Watch` section in readme file ?

glensc:
Not really:
- https://github.com/Taxel/PlexTraktSync/pull/480#issuecomment-927171480

also, I'd like to address [plex server restart issue](https://github.com/Taxel/PlexTraktSync/pull/502#issuecomment-927153875) before making this public.

glensc:
Server restart handler:
- https://github.com/Taxel/PlexTraktSync/pull/508

PR #507, opened 2021-09-26
Add systemd instructions for watch command

agneevX:


README.md
```diff
@@ -168,6 +168,25 @@ and scrobble plays.
 
 [plex-scrobbler]: https://blog.trakt.tv/plex-scrobbler-52db9b016ead
 
+Create a systemd daemon so that it scrobbles automatically in the background:
```
glensc, line 171:
```suggestion
Create a systemd unit so that it scrobbles automatically in the background:
```

agneevX:
It seems that `python3 -m plex_trakt_sync watch` isn't correct?

Shouldn't it be `./plex_trakt_sync.sh watch`?

glensc:
Explain your claim!

agneevX:
Ok. The main entrypoint to using PlexTraktSync is the shell script right? So it's only fair that the python executable not be called directly.

glensc:
@agneevX is this ready for merge?

agneevX:
Yes of course.

glensc:
@agneevX why "of course"? you had a question, and you never explained or showed what error you encountered with readme, nor any further feedback.

PR #508, opened 2021-09-26
Feature: Add Error event listener to WatchStateUpdater

glensc:
This allows WatchStateUpdater to reset internal cache on server restarts.


glensc:
I've submitted a proposal to get real errors from AlertListener:
- https://github.com/pkkid/python-plexapi/pull/834

PR #512, opened 2021-09-27
Use InquirerPy for user input

sirloinofbeef:
Simplified user input using InquirerPy:
- Server selection _(selectable list)_
image

- Managed user Y/N _(made redundant by below)_
- Managed user selection _(selectable list, including default main username)_
image

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -19,6 +19,7 @@ tqdm = "==4.62.3"
 trakt = "==3.2.1"
 urllib3 = ">=1.26.6"
 websocket-client = "==1.2.1"
+inquirerpy = "==0.2.4"
 
 [requires]
 python_version = "3"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "2fc0672e6173b5522e140185b7ff9cbe53d3bcd17ae80acb09472043d90dc686"
+            "sha256": "915a6023706c21958d1c7ed03c3cb0e14659217c0d0de52ca417209fba1fc861"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -21,6 +21,7 @@
                 "sha256:7d5d0167b2b1ba821647616af46a749d1c653740dd0d2415100fe26e27afdf41",
                 "sha256:a841dacd6b99318a741b166adb07e19ee71a274450e68237b4650ca1055ab128"
             ],
+            "index": "pypi",
             "version": "==1.4.4"
         },
         "attrs": {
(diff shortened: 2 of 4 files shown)


glensc:
I don't see the point in adding another dependency.

the password wasn't shown already with the current implementation.

sirloinofbeef:
Imho having * appear makes input easier / more intuitive. The main bonus though comes with the selection input for servers / users. Makes it impossible for users to enter typos or incorrect data. The managed user Y/N and selection inputs are also streamlined into a single input.

If this is not a direction wanted for the script though, that's fine, and / or I can look into coding something which doesn't require the dependancy.

glensc:
`having * appear` only for windows users, unix users know that password input never shows anything. and you should add screenshots

glensc:
We already have `rich` library in dependencies:
- https://github.com/willmcgugan/rich

sirloinofbeef:
> We already have `rich` library in dependencies:
> 
> * https://github.com/willmcgugan/rich

Afaik `rich` is for RTF and doesn't offer the selection input which this change is making.

glensc:
a4d0bbd: you misunderstood me, it's a mindset of windows users that password must be shown as stars. other users consider that that's leaking password info if you show how long is your password while you're typing and someone peeks over your shoulders. I never said you should make such changewindows only. what I said, that rather using yet another framework as dependency, use existing ones. we have click, we have rich already as deps.

it's my opinion (other maintainers might think otherwise). also, I won't support any windows specific code at all. ever. python is platform-independent and should stay that way here.

glensc:
Server selection: you lost information of server details that were printed before

sirloinofbeef:
> Screenshots should go to PR body
Fixed

> Server selection: you lost information of server details that were printed before
No, that information is still there. The selectable list appears directly after it as per the current server selection input.

> that's leaking password info
Completely understand. The idea of the password * is to make things more user friendly for a susbset of users. (Windows users). If this presents too much of a concern then I'm happy to remove it, particularly if it distracts from the main purpose of this push, selectable server / user input options rather than typed (which are more susceptible to to user error e.g. typos). I only tacked on the password input alteration as the dependency happened to support it on top of the selectable input options.

> I won't support any windows specific code
That's fair.

glensc:
1. please remove platform specific code. use same style for all platforms
2. change "main user" to show actual login of main account. maybe add dash separator between it and managed users?
3. your screenshot shows only server names, doesn't show server IP, URL, last seen, server version

```
- ONEPLUS A5000: [Last seen: 2021-09-27 21:54:39, Server: Plex for Android (Mobile)/8.23.1.28053 on ONEPLUS A5000: Android/10]
    http://192.168.0.30:32500
- Vulpes: [Last seen: 2021-09-28 08:26:17, Server: Plex Media Server/1.22.3.4392-d7c624def on Docker Container (LinuxServer.io): Linux/4.14.134-boot2docker]
    https://192-168-0-xxx.plex.direct:32400
    https://192-168-0-xxx.plex.direct:32400
```

glensc:
default password via commandline/env is lost: `default=password`

sirloinofbeef:
> 1. please remove platform specific code. use same style for all platforms

Done

> 2. change "main user" to show actual login of main account. maybe add dash separator between it and managed users?

I'll lookup how to do this and implement.

> 3. your screenshot shows only server names, doesn't show server IP, URL, last seen, server version

Screenshot updated to show positioning in flow.

sirloinofbeef:
> default password via commandline/env is lost: `default=password`

Double check and it IS there as per the original code.

glensc:
> > default password via commandline/env is lost: `default=password`
> 
> Double checked and it IS there as per the original code.

yes, it's back once you removed platform-specific code.

PR #520, opened 2021-09-29
Use spinner from rich for trakt lists cache status

glensc:
[asciicast](https://asciinema.org/a/6agVO8dq5Iuvqpm8LTXybjWkQ)


twolaw:
Are you sure `lists` is what defines the n items cached in `console.log(f"Cached {n} Trakt lists")` ?
It doesn't refer to user liked lists ? Maybe there could be confusion.

glensc:
@twolaw I know not all are lists. suggest a better name, end-user doesn't care anyway.

twolaw:
Previous code didn't say a wrong number of lists.
I suggest to remove the number.

glensc:
The number is not inaccurate

twolaw:
It is useless and appended to a wrong unit.

twolaw:
Calm down, I'm not complaining I'm suggesting with arguments. Something we always do here.

> suggest something useful.

I already made a suggestion: to remove the n number. And I explained why.

glensc:
Remove `n` is not a valid suggestion. Not using the variable results in dead code.

twolaw:
I mean in the `console.log(f"Cached {n} Trakt lists")`
Becomes : `console.log(f"Trakt lists cached")`

glensc:
And I mean removing `{n}` printing makes assigning `n` variable dead code.

PR #523, opened 2021-10-03
Add multi arch support for Docker images

### LINKED ISSUE #522: Add multi arch support
I've tried running the container on `arm64` but I get an exec error. 

It seems to be `amd64` only. Please add support for other architectures.

agneevX:
Add support for arm64, `armhf`/`armv7`. Close #522


glensc:
does not work:
- https://github.com/Taxel/PlexTraktSync/issues/522#issuecomment-932916268

PR #536, opened 2021-10-12
Generate requirements.txt from Pipenv

glensc:
This should solve dependabot problem of updating only Pipfile.

As I understand, if requirements.txt is generated this way, then dependabot is able to update it as well.

See:
- https://github.com/dependabot/feedback/issues/435#issuecomment-483259640

Also:
- https://github.com/Taxel/PlexTraktSync/pull/400#issuecomment-888088343 cc @chrillep
- https://github.com/Taxel/PlexTraktSync/pull/397#issuecomment-887637015 cc @twolaw
- https://github.com/dependabot/feedback/issues/435
- https://github.com/dependabot/dependabot-core/issues/1946


chrillep:
yes after reading the comments again. I misunderstood what generating through pipen would do for dependabot. so good catch 😊

Issue #591, opened 2021-11-07
Shows Sync: Internal Server Error for plextraktsync.trakt_api.search_by_id()


fnwc:
### The problem

Using tag 0.16.1.

Syncing movies seems to work fine, 

Using `python3 -m plextraktsync sync --sync=shows` or even  `python3 -m plextraktsync sync --sync=tv` I get the following error:

![image](https://user-images.githubusercontent.com/7389518/140664848-cec86009-9387-4ae4-bff2-f0ab8e1890dc.png)

Running with `"log_debug_messages": true`, see error trace.

### Error trace

```text
2021-11-07 14:39:24,806 DEBUG[trakt.core]:get: https://api-v2launch.trakt.tv/search/tvdb/3610949641284564056?type=show
2021-11-07 14:39:24,806 DEBUG[trakt.core]:headers: {'Content-Type': 'application/json', 'trakt-api-version': '2', 'trakt-api-key': 'removed', 'Authorization': 'Bearer removed'}
2021-11-07 14:39:24,806 DEBUG[trakt.core]:method, url :: get, https://api-v2launch.trakt.tv/search/tvdb/3610949641284564056?type=show
2021-11-07 14:39:25,132 DEBUG[urllib3.connectionpool]:https://api-v2launch.trakt.tv:443 "GET /search/tvdb/3610949641284564056?type=show HTTP/1.1" 500 None
2021-11-07 14:39:25,132 DEBUG[requests_cache.session]:Pre-cache checks for response from https://api-v2launch.trakt.tv/search/tvdb/3610949641284564056?type=show: {'disabled cache': False, 'disabled method': False, 'disabled status': True, 'disabled by filter': False, 'disabled by headers or expiration params': False}
2021-11-07 14:39:25,132 DEBUG[requests_cache.session]:Skipping cache write for URL: https://api-v2launch.trakt.tv/search/tvdb/3610949641284564056?type=show
2021-11-07 14:39:25,132 DEBUG[trakt.core]:RESPONSE [get] (https://api-v2launch.trakt.tv/search/tvdb/3610949641284564056?type=show): <Response [500]>
2021-11-07 14:39:25,132 WARNING[PlexTraktSync]:Internal Server Error for plextraktsync.trakt_api.search_by_id(), retrying after 1 seconds (try: 1/5)
2021-11-07 14:39:26,145 DEBUG[trakt.core]:get: https://api-v2launch.trakt.tv/search/tvdb/3610949641284564056?type=show
[...]

PR #602, opened 2021-11-08
Fix: Skip search_by_id with invalid input

glensc:
`search_by_id` takes guid.id as intput, and that must be numeric, or with `tt` prefix (imdb), with too long input, trakt gives out 500 error.

Refs @twolaw  note:
- https://github.com/Taxel/PlexTraktSync/issues/591#issuecomment-963553557

### DIFF
--- a/plextraktsync/trakt_api.py
+++ b/plextraktsync/trakt_api.py
@@ -226,6 +226,11 @@ class TraktApi:
             logger.debug("tvdb does not support movie provider")
             return None
 
+        if not self.valid_trakt_id(media_id):
+            logger.error(f"Ignoring invalid id: '{media_id}'")
+
+            return None
+
         search = trakt.sync.search_by_id(media_id, id_type=id_type, media_type=media_type)
         # look for the first wanted type in the results
         # NOTE: this is not needed, kept around for caution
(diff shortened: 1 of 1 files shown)


twolaw:
Are you sure isnumeric() is a good test ?

`AttributeError: 'int' object has no attribute 'isnumeric'`

```
> a='3610949641284564056'
> a.isnumeric()
True
```

glensc:
Perhaps switch type(int) to be first. altho i think no code calls media_id as integer, always string, and also the imdb id is with tt prefix

glensc:
removed int test, as guid.id is always string. also allowed imdb tt prefix

glensc:
ok, the numeric test won't help for becase it would pass the test as being valid:

https://github.com/Taxel/PlexTraktSync/issues/591#issuecomment-963622787:
```
<tvdb:3610949641284564056:<Show:705490:60-Days-In:-Narcolan>>
```

so we have to test that length is not 19?

glensc:
@twolaw ok now?

twolaw:
Trakt real limit seems to be about 10 digits.
I suggest to lower the length test to 12 by safety.

twolaw:
> `search_by_id` input must be  numeric, or with `tt` prefix (imdb), not doing so makes trakt give 500 error.

That's not correct. Trakt can deal with any letters in ids. Only length matters.

glensc:
> > `search_by_id` input must be  numeric, or with `tt` prefix (imdb), not doing so makes trakt give 500 error.
> 
> That's not correct. Trakt can deal with any letters in ids. Only bad length raises a 500 error.

Point was that the method is in our library, it's searching by id. and ids are numeric with imdb exception. 

For searching by slug, I recommend creating another specific method in our library, search_by_slug.

I'll change the limit to 12 and merge. but you should add a link to your information where you took this information.

twolaw:
Looks good to me 👍 

The 10 digit comes from empiric experience. I tried to send trakt requests with:
- 9 digit : always return code 200
- 10 digit : some return code 200 and some return code 500
- 11 digit : always return code 500

PR #604, opened 2021-11-11
Feature: Deep merge config.defaults.json

glensc:
This helps in cases when config.default.json will have new key under section that already has a value:

// config.json
```json
    "logging": {
        "append": false
    },
```

// config.default.json
```json
    "logging": {
        "append": true,
        "filename": "plextraktsync.log"
    },
```

without this changeset, the final config would be:

```json
    "logging": {
        "append": false
    },
```

but needs to be:

```json
    "logging": {
        "append": false,
        "filename": "plextraktsync.log"
    },
```

### DIFF
--- a/plextraktsync/config.py
+++ b/plextraktsync/config.py
@@ -63,7 +63,7 @@ class Config(dict):
                 fp.write(json.dumps(defaults, indent=4))
 
         config = self.load_json(self.config_file)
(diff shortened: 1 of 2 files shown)


glensc:
Simple testing worked, but unit tests should be added/updated.

Issue #619, opened 2021-11-23
Play count gets incremented on Plex for certain movies.


Rouzax:
### The problem

Some movies keep getting an additional play count at the time the plexytraktsync script is run

```
INFO: Marking as watched in Plex: <tmdb:259991:<Movie:121394:Ernst,-Bobbie-en-de->>
INFO: Marking as watched in Plex: <tmdb:324868:<Movie:121395:Ernst,-Bobbie-en-de->>
INFO: Marking as watched in Plex: <tmdb:503676:<Movie:121400:Ernst,-Bobbie-en-de->>
INFO: Marking as watched in Plex: <tmdb:588016:<Movie:121399:Ernst,-Bobbie-en-de->>
INFO: Marking as watched in Plex: <tmdb:508570:<Movie:128712:The-One-and-Only-Iva>>
INFO: Marking as watched in Plex: <tmdb:271525:<Movie:116879:Sprookjesboom-de-Mus>>
INFO: Marking as watched in Plex: <tmdb:391488:<Movie:116878:Sprookjesboom-de-Mus>>
INFO: Marking as watched in Plex: <tmdb:266297:<Movie:116880:Woezel-En-Pip---Alle>>
```

### Error trace

```text
2021-11-23 10:32:12,247 DEBUG[plexapi]:GET https://192-168-2-13.50f5366c69c94499b5515522c37d9ec1.plex.direct:32400/library/metadata/4251?checkFiles=1&includeAllConcerts=1&includeBandwidths=1&includeChapters=1&includeChildren=1&includeConcerts=1&includeExternalMedia=1&includeExtras=1&includeFields=thumbBlurHash%2CartBlurHash&includeGeolocation=1&includeLoudnessRamps=1&includeMarkers=1&includeOnDeck=1&includePopularLeaves=1&includePreferences=1&includeRelated=1&includeRelatedCount=1&includeReviews=1&includeStations=1
[...]

PR #640, opened 2021-12-12
Fix: restore python 3.8 support

glensc:
The typing was incorrect anyway:
```py
      list[Movie, Show, Episode]
```

The values are actually strings, not objects.

refs:
- https://github.com/Taxel/PlexTraktSync/issues/619#issuecomment-991923551
- https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

### DIFF
--- a/plextraktsync/walker.py
+++ b/plextraktsync/walker.py
@@ -91,7 +91,7 @@ class WalkPlanner:
         if not self.config.id:
             return [None, None, None]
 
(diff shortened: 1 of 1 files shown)


glensc:
Confirmed:
- https://github.com/Taxel/PlexTraktSync/issues/619#issuecomment-991926922

cc @animosity22

PR #587, opened 2021-11-06
Refactor: Rename "version" command to "info"

glensc:
Rename "version" command to "info" and put all info printing to that command.

Besides putting all useful info to same place,
this improves privacy when people paste sync command output to bug reports.

PR #658, opened 2021-12-19
Fix: correct info command in issue template

agneevX:


glensc:
Command was renamed here:
- https://github.com/Taxel/PlexTraktSync/pull/587

PR #664, opened 2021-12-19
Feature: Use plextraktsync.log as log filename

glensc:
This enables append=true to keep all logs, therefore having last_update as log filename is no longer appropriate.

Also, moved all logging options under "logging" in config file.


glensc:
This change been on my mind for very long time!

Issue #631, opened 2021-12-08
Media play event syncs only after cache DB is recreated


agneevX:
### The problem

Sometimes media play events won't sync until the cache DB is recreated.

```py
# Initial sync
ubuntu@oc-hyd1 ~ [1]> ./scripts/trakt_run.sh sync --sync movies
INFO: PlexTraktSync [0.17.0]
WARNING: No response from http://10.0.0.11:32400, fallback to http://falcon.tailnet-d10c.ts.net:32400
Sync Movie sections: [<PlexLibrarySection:movie:Movies>, <PlexLibrarySection:movie:Movies (4K)>]
Processing Movies 100% ━━━━━━━━━━━╸ 1,255/1,258  [ 0:00:42 < 0:00:01 , 33 it/s ]
INFO: Movies processed in 42.9 seconds
Processing Movies (4K)  96% ━━━━━━━━━━━━  91/95  [ 0:00:03 < 0:00:01 , 54 it/s ]
INFO: Movies (4K) processed in 3.4 seconds
INFO: Updated plex watchlist in 0.0 seconds
INFO: Completed full sync in 46.3 seconds

ubuntu@oc-hyd1 ~> rm /opt/appdata/plextraktsync/trakt_cache.sqlite

# Sync done for specific library only
ubuntu@oc-hyd1 ~> ./scripts/trakt_run.sh sync --library "Movies (4K)"
INFO: PlexTraktSync [0.17.0]
WARNING: No response from http://10.0.0.11:32400, fallback to http://falcon.tailnet-d10c.ts.net:32400
Sync Movie sections: [<PlexLibrarySection:movie:Movies (4K)>]
INFO: Marking as watched in Trakt: <tmdb:512195:<Movie:81431:Red-Notice>>
Processing Movies (4K) 100% ━━━━━━━━━━━━━━ 95/95  [ 0:02:37 < 0:00:00 , 1 it/s ]
INFO: Movies (4K) processed in 2 min 37.6 seconds
INFO: Updated plex watchlist in 0.0 seconds
INFO: Completed full sync in 2 min 37.6 seconds
```

```sh
# trakt_run.sh
#!/bin/bash
set -e

command () {
  docker run --rm --name pts-sync \
  --hostname pts-sync \
  -u "1001:1001" \
  -v /opt/appdata/plextraktsync:/app/config \
  ghcr.io/taxel/plextraktsync "$@"
}

command "$@"
```

[...]

PR #666, opened 2021-12-19
Fix: Prevent rating key access filling cache with stale data

glensc:
This may call fetch by plex and it may be cached
or later code could prevent last seen date being incorrect

Probably fixes https://github.com/Taxel/PlexTraktSync/issues/631 and fixes https://github.com/Taxel/PlexTraktSync/issues/619


glensc:
Evil PR number to fix rather Evil bug :D

glensc:
This super kills performance, for each movie:

```
2021-12-19 19:15:56,853 DEBUG[plexapi]:Reloading Movie 'xxx' for attr 'userRating'
```

glensc:
Quick workaround: Disable ratings syncing.

PR #668, opened 2021-12-19
Refactor: Fetch Plex userRatings in a separate call

glensc:
Since rating fetching for all items is super slow:
- https://github.com/Taxel/PlexTraktSync/pull/666#issuecomment-997428206

add method to find Plex items that have rating value present.


glensc:
Use rating search results to pre-fill rating method value

PR #684, opened 2022-01-02
Add logging to README.md and Re-structure

stevietv:
I went ahead and added the information about the new logging options in config.json to the README.md

At the same time I restructured it somewhat, grouping the installation methods together (instead of having windows and unraid spread through). and added a TOC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,29 @@ unlike the Plex app provided by Trakt.
 
 Note: The PyTrakt API keys are not stored securely, so if you do not want to have a file containing those on your harddrive, you can not use this project.
 
+- [Plex-Trakt-Sync](#plex-trakt-sync)
+  - [Features](#features)
+  - [Pre-requisites](#pre-requisites)
+  - [Installation](#installation)
+    - [pipx](#pipx)
+    - [GitHub](#github)
+      - [GitHub download](#github-download)
+      - [GitHub clone](#github-clone)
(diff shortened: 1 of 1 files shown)


glensc:
I  generally prefer atomic commits:
- https://www.freshconsulting.com/insights/blog/atomic-commits/

meaning each change kind be a different commit.

also, do cross link things if you come to this PR via discussion or an issue. just edit the PR body.

stevietv:
noted on the atomic commits. cheers!

glensc:
actually, I would like to remove git and github installation methods and keep them as developer options (contributing.md), but I don't know why people do install that way.

stevietv:
I chose to install via a git clone as its now super easy to update through a terminal. just fetch the tags and checkout the latest.
No need to download, extract and clean up zips.

glensc:
but pipx is just single command and with git you also need to keep dependencies up to date. and without virtualenv and other packages it may even not be possible.

stevietv:
makes sense, want me to move it over to contributing.md?

Maybe a little OT: I take it to migrate from a git clone to a pipx install then the `.env`, `config.json` and `trakt_cache.sqlite` would need to be moved over to the new appdir

glensc:
yes, confs and cache need to be moved manually. or as they are missing in new location interactive configure will ask credentials again.

sure, move, but leave a link to readme that git/github things are for development and links to contributing.

PR #706, opened 2022-01-10
Revert "Bump requests-cache from 0.8.1 to 0.9.0 (#686)"

glensc:
This reverts commit b62dc4555e95cde2b90e8f7a9f77ef510e274f2a from #686.

See https://github.com/reclosedev/requests-cache/issues/494

The warnings will confuse users of this app, and I'm not interested in explaining to them that they can ignore it.

Ideally, this could be investigated further and perhaps report to trakt.tv that they provide a non-json response for JSON content type.


glensc:
I'm not sure how to tell @dependabot to stop updating this dependency (since I'm downgrading here) if the PR was already merged. but gave it a try anyway:
- https://github.com/Taxel/PlexTraktSync/pull/686#issuecomment-1008919057

PR #716, opened 2022-01-12
Refactor: Use TVShow.episodes property for episodes trakt id lookup

glensc:
Requires:
- https://github.com/moogar0880/PyTrakt/pull/176

### DIFF
--- a/plextraktsync/pytrakt_extensions.py
+++ b/plextraktsync/pytrakt_extensions.py
@@ -2,22 +2,6 @@ from trakt.core import get
 from trakt.tv import TVEpisode
 
 
-@get
-def lookup_table(show):
-    # returns all seasons and episodes with one single call
-    data = yield "shows/{}/seasons?extended=episodes".format(show.trakt)
-    retVal = {}
-    for season in data:
-        eps = {}
-        if "episodes" in season.keys():
-            for episode in season["episodes"]:
-                eps[episode["number"]] = LazyEpisode(
-                    show, season["number"], episode["number"], episode["ids"]
-                )
-        retVal[season["number"]] = eps
-    yield retVal
-
-
 class LazyEpisode:
     def __init__(self, show, season, number, ids):
         self.show = show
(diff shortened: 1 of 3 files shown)


twolaw:
⚠️ this PR breaks `TraktLookup` functionnality, do not release as is

glensc:
perhaps explain the problem? steps to reproduce?

twolaw:
Just sync a show (i tested Attack on Titan) and see that `find_by_guid` is called for each episode.
With `0.23.3` there is no `find_by_guid` called because the lookup table works.

glensc:
Unable to reproduce, which season/episode is failing? 

I added s01e01 test:
- https://github.com/Taxel/PlexTraktSync/pull/1129

just to be sure:
1. you installed required dependencies?
2. you cleared cache so it's not related to cache problem?

twolaw:
I found the problem.

The seasons() from trakt.tv.py uses `extract_ids()` to flatten the ids dict so the `te.ids.get(guid.provider)` doesn't work anymore, it always return None.

`te.ids.get(guid.provider)` needs to be replaced by `getattr(te, guid.provider)`

glensc:
Found the problem:
- [`17abf87` (#1129)](https://github.com/Taxel/PlexTraktSync/pull/1129/commits/17abf87ca610164ce699bc26a11afa9f6e42ef30)

glensc:
So, perhaps the PR was incomplete:
- https://github.com/moogar0880/PyTrakt/pull/176/files

should add extract_ids to the `seasons()` method?

but you seem to claim opposite it does extract_ids while I see it doesn't do extract_ids

twolaw:
I think we should use `extract_ids()` everywhere in pytrakt to be consistent with moogar0880 code.

glensc:
> I think we should use extract_ids() everywhere in pytrakt to be consistent with moogar0880 code.

agree, as without that, properties like `ids` would return `None` for all providers.  `extract_ids` missing, is bug.

PR #726, opened 2022-01-14
Feature: Move memoize from trakt_api to media class

glensc:
remove `@memoize` from:
- trakt_api::collected -> moved to media::collected
- trakt_api::lookup -> moved to media::seasons

so the cache is per episode/movie, not in the global traki_api class

see usage report:
- https://github.com/Taxel/PlexTraktSync/issues/431#issuecomment-1013444604


glensc:
no feedback, so i'll just merge and release. also works for me

PR #727, opened 2022-01-15
Performance: Drop memoize of find_by_guid

glensc:
It's not useful for sync run and only consumes memory for cache that
nobody needs:
- https://github.com/Taxel/PlexTraktSync/issues/431#issuecomment-1013709208

originally added in ef991ea66b520bf243cfd0774afb379c5a1e6442

Fixes https://github.com/Taxel/PlexTraktSync/issues/431


glensc:
Merging:
- https://github.com/Taxel/PlexTraktSync/issues/431#issuecomment-1013746446

PR #738, opened 2022-01-20
Refactor: Lazy import commands

glensc:
This prevents all commands being imported just to figure out their usage.

This has minor speedup for setting up the application due not importing commands that aren't executed in the current run.

### DIFF
--- a/plextraktsync/cli.py
+++ b/plextraktsync/cli.py
@@ -1,7 +1,6 @@
 from functools import wraps
 import click
 
-from plextraktsync.commands.clear_collections import clear_collections
 from plextraktsync.commands.info import info
 from plextraktsync.commands.inspect import inspect
 from plextraktsync.commands.login import login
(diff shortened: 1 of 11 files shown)


AndreiArdelean1:
A few issues here:  
unmatched no longer works
```
Usage: python -m plextraktsync [OPTIONS] COMMAND [ARGS]...`
Try 'python -m plextraktsync --help' for help.
Error: No such command 'unmatched'.
```

and 

And a new `wrap` command?
```
Usage: python -m plextraktsync [OPTIONS] COMMAND [ARGS]...
  Plex-Trakt-Sync is a two-way-sync between trakt.tv and Plex Media Server
Options:
  --help  Show this message and exit.

:
  cache              Manage and analyze Requests Cache.
  clear-collections  Clear Movies and Shows collections in Trakt
  info               Print application and environment version info
  inspect            Inspect details of an object
  login              Log in to Plex and Trakt if needed
  plex-login         Log in to Plex Account to obtain Access Token.
  sync               Perform sync between Plex and Trakt
  trakt-login        Log in to Trakt Account to obtain Access Token.
  watch              Listen to events from Plex
  wrap               List media that has no match in Trakt or Plex
```

glensc:
This is fixed in 62857499, released now as 0.18.15

PR #605, opened 2021-11-11
Add config to configure trakt auto watch threshold

glensc:
adds logic to do pause if watch progress is below threshold:

```py
    def stop(self, progress: float):
        if progress >= self.threshold:
            self.scrobbler.stop()
        else:
            self.scrobbler.pause()
```

Closes https://github.com/Taxel/PlexTraktSync/issues/603

Needs to be merged first:
- https://github.com/Taxel/PlexTraktSync/pull/604

PR #739, opened 2022-01-21
Feature: Marking as Watched in Trakt in watch command

Hollander-1908:
Marking as watched in Trakt when watched in Plex but not marked as watched in Trakt yet.
This will help when manually select "Mark as Played" in Plex without playing the video.

PR #741, opened 2022-01-22
Mark as watched if above scrobble threshold

Dalethium:
I am not sure if this is intended for the project, but this is what I really wanted out of this project so I figured I'd at least put this here.

refs:
- https://github.com/Taxel/PlexTraktSync/pull/605

### DIFF
--- a/plextraktsync/commands/watch.py
+++ b/plextraktsync/commands/watch.py
@@ -47,6 +47,7 @@ class WatchStateUpdater:
         self.plex = plex
         self.trakt = trakt
         self.mf = mf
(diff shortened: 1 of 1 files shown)


Dalethium:
This is also what I was under the impression that #739 did, and most of what I changed was based off that.

glensc:
This is not needed. trakt.tv does this automatically.

Hollander-1908:
I think #750 is a better approach.
Now it will only mark as watched on Trakt when Plex decides it's watched completely.
With this method it will both trigger Plex and Trakt to mark it as watched on whatever treshold you set it on.

PR #750, opened 2022-01-24
[watch] Mark as watched on Plex if above scrobble threshold

Hollander-1908:
Mark as watched on Plex when going over the percentage treshold.
This will also mark it as watched on Trakt automatically.

I think the default treshold should be higher than 80 though. 
Most of the episodes has short end credits, I think 90 or even 95 should be better for episodes.
80 should be helpfull for movies because of the long end credits.

### DIFF
--- a/plextraktsync/commands/watch.py
+++ b/plextraktsync/commands/watch.py
@@ -121,7 +121,7 @@ class WatchStateUpdater:
         movie = m.plex.item
         percent = m.plex.watch_progress(event.view_offset)
         
-        if percent >= self.scrobblers.threshold:
+        if percent >= self.threshold:
             m.mark_watched_plex()
 
         self.logger.info(f"{movie}: {percent:.6F}% Watched: {movie.isWatched}, LastViewed: {movie.lastViewedAt}")
--- a/plextraktsync/commands/watch.py
+++ b/plextraktsync/commands/watch.py
@@ -120,6 +120,9 @@ class WatchStateUpdater:
 
         movie = m.plex.item
         percent = m.plex.watch_progress(event.view_offset)
+        
+        if percent >= self.scrobblers.threshold:
+            m.mark_watched_plex()
 
         self.logger.info(f"{movie}: {percent:.6F}% Watched: {movie.isWatched}, LastViewed: {movie.lastViewedAt}")
         self.scrobble(m, percent, event)


plextraktsync/commands/watch.py
```diff
@@ -120,6 +120,9 @@ def on_play(self, event: PlaySessionStateNotification):
 
         movie = m.plex.item
         percent = m.plex.watch_progress(event.view_offset)
+        
+        if percent >= self.scrobblers.threshold:
```
glensc, line 124:
```suggestion
        if percent >= self.threshold:
```

glensc:
correct PR title to indicate you changed watch command

plextraktsync/commands/watch.py
Hollander-1908: I guess both will work but sure `self.treshold` is fine :)

Hollander-1908:
Changed PR title.

glensc:
Still, title lacks detail that this affects the watch not sync command.

Hollander-1908:
> Still, title lacks detail that this affects the watch not sync command.

Is it fine now?

agneevX:
> Mark as watched on Plex when going over the percentage treshold.

So this changes the logic that Plex uses? Just why?

twolaw:
> I think the default treshold should be higher than 80 though.

I disagree.
Ok to let the user change the threshold in config file but the default should remain as Plex default.

plextraktsync/commands/watch.py
glensc: you have not tested your changes with f12ccd2a7e36ab4cc11e6ea0e745d0dad9f9fb91 commit, the self.threshold is not initialized in this branch like the other pr you took changes from.

glensc:
> > I think the default treshold should be higher than 80 though.
> 
> I disagree. Ok to let the user change the threshold in config file but the default should remain as Plex default. The script is not supposed to change Plex behaviour until user opt-in.

It may make sense to have a different threshold for movies and series. I'm willing to implement it if someone makes feature request issue :) but nevertheless. this would be trakt threshold.

glensc:
Perhaps this should be implemented differently. If media play is stopped. let trakt.tv handle played state and then perform sync of trakt->plex. however may need to delay the action to allow trakt.tv site to be slow.

so, I understand the problem that is being tried to solve is that user has configured threshold to 70%, now pressing stop will make as watched in trakt. but on plex side the item is not marked as played as plex has own 80% logic, this could go in sync if you run sync for that specific media item, but watch command doesn't do that.

SwAgLaNdErr:
> Perhaps this should be implemented differently. If media play is stopped. let trakt.tv handle played state and then perform sync of trakt->plex. however may need to delay the action to allow trakt.tv site to be slow.
> 
> so, I understand the problem that is being tried to solve is that user has configured threshold to 70%, now pressing stop will make as watched in trakt. but on plex side the item is not marked as played as plex has own 80% logic, this could go in sync if you run sync for that specific media item, but watch command doesn't do that.

I understand what you're saying. But isn't the watch command there because it will reduce API hits in the first place? If it can mark it as watched in Trakt when you have played something while using watch command you don't need to do a sync every *minutes/*hour, right? I watched two episodes of two different shows last night and it only marked one episode as watched in Trakt while both are watched in Plex. So or there's some kind of bug or I don't understand the purpose of the watch command lol

I'll add an feature request for movie and episode treshold :)

Hollander-1908:
> Perhaps this should be implemented differently. If media play is stopped. let trakt.tv handle played state and then perform sync of trakt->plex. however may need to delay the action to allow trakt.tv site to be slow.
> 
> so, I understand the problem that is being tried to solve is that user has configured threshold to 70%, now pressing stop will make as watched in trakt. but on plex side the item is not marked as played as plex has own 80% logic, this could go in sync if you run sync for that specific media item, but watch command doesn't do that.

I understand what you're saying. But isn't the watch command there because it will reduce API hits in the first place? If it can mark it as watched in Trakt when you have played something while using watch command you don't need to do a sync every *minutes/*hour, right? I watched two episodes of two different shows last night and it only marked one episode as watched in Trakt while both are watched in Plex. So or there's some kind of bug or I don't understand the purpose of the watch command lol

I'll add an feature request for movie and episode treshold :)

Hollander-1908:
Added feature request #756

Issue #816, opened 2022-02-20
PlexTraktSync won't connect server with new dynamic IP


fqx:
### The problem

When Plex server is on a dynamic IP, PlexTraktSync won't connect server with latest IP/URL.

### Error trace

```text
INFO: PlexTraktSync [0.18.20-6-g781fe5c_main]
ERROR: Plex connection error: HTTPSConnectionPool(host='116-21-1-66.9686607bf3264d53ac73839a66e89920.plex.direct', port=32400): Max retries exceeded with url: / (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7fe9e8ea1db0>, 'Connection to 116-21-1-66.9686607bf3264d53ac73839a66e89920.plex.direct timed out. (connect timeout=30)')), fallback url http://172.17.0.1:32400 didn't respond either.
Plex connection error: HTTPSConnectionPool(host='116-21-1-66.9686607bf3264d53ac73839a66e89920.plex.direct', port=32400): Max retries exceeded with url: / (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7fe9e8ea1db0>, 'Connection to 116-21-1-66.9686607bf3264d53ac73839a66e89920.plex.direct timed out. (connect timeout=30)')), fallback url http://172.17.0.1:32400 didn't respond either.
```

### Expected behavior

Connect to 113-111-63-251.9686607bf3264d53ac73839a66e89920.plex.direct:32400

### Workarounds

Maybe loose secure connection requirements could work, but I won't try, it is not safe.

### Install method

docker-compose

### Version

0.18.20-6-g781fe5c_main

### Python Version

3.10.0

### Operating System and Version

Docker

glensc:
FYI: we don't support issues using installation from branch: 0.18.20-6-g781fe5c_main, use released versions. and whoever instructed you to use branch main, complain to them as well.

glensc:
> Workarounds
>
> Maybe loose secure connection requirements could work, but I won't try, it is not safe.

you can use `plex-login` subcommand to find new IP each time it changes to find the new ip

glensc:
Probably alternative would be to use MyPlex account and retrieve server address dynamically each time. Slower, but should catch the updates of changing server addresses.

fqx:
> Probably alternative would be to use MyPlex account and retrieve server address dynamically each time. Slower, but should catch the updates of changing server addresses.

how do you do that?

[...]

PR #817, opened 2022-02-21
Rename Plex variable fallbackurl to localurl

twolaw:
In `.env` file, I suggest renaming the variable `PLEX_FALLBACKURL` to `PLEX_LOCALURL` so that **users understand and change this URL by themself** with a specific message in error logs.

______
After that, the local IP address auto-finding will have to be improved for docker users because 172.17.0.x doesn't look useful (see #775 and #816)

### DIFF
--- a/plextraktsync/plex_server.py
+++ b/plextraktsync/plex_server.py
@@ -23,7 +23,7 @@ def _get_plex_server():
     CONFIG = factory.config()
     plex_token = CONFIG["PLEX_TOKEN"]
     plex_baseurl = CONFIG["PLEX_BASEURL"]
-    plex_fallbackurl = CONFIG["PLEX_FALLBACKURL"]
+    plex_localurl = CONFIG["PLEX_LOCALURL"]
     if plex_token == "-":
         plex_token = ""
     server = None
@@ -37,13 +37,13 @@ def _get_plex_server():
     # if connection fails, it will try :
     # 1. url expected by new ssl certificate
     # 2. url without ssl
-    # 3. fallback url (localhost)
+    # 3. local url (localhost)
 
     try:
         server = PlexServer(token=plex_token, baseurl=plex_baseurl)
     except plexapi.server.requests.exceptions.SSLError as e:
-        m = "Plex connection error: {}, fallback url {} didn't respond either.".format(
-            str(e), plex_fallbackurl
+        m = "Plex connection error: {}, local url {} didn't respond either.".format(
+            str(e), plex_localurl
         )
         excep_msg = str(e.__context__)
         if "doesn't match '*." in excep_msg:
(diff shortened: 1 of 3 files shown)


plextraktsync/plex_server.py
```diff
@@ -58,7 +60,9 @@ def _get_plex_server():
                 # save new url to .env
                 CONFIG["PLEX_TOKEN"] = plex_token
                 CONFIG["PLEX_BASEURL"] = new_plex_baseurl
-                CONFIG["PLEX_FALLBACKURL"] = plex_fallbackurl
+                plex_localurl = CONFIG.get("PLEX_LOCALURL")
```
glensc, line 63: have you tested does `CONFIG.get("PLEX_LOCALURL")` thow if key is missing? maybe need `CONFIG.get("PLEX_LOCALURL", None)`

glensc:
In my mind there should be a change in config.py as well, adding PLEX_LOCALURL environment extraction.

I would probably make fallback to previous variable in config class, not in the outside (plex-server.py) doing similar checks multiple times.

glensc:
also, atomic commits. renaming things and adding fallback should have been two separate commits.

glensc:
also added "docs needed" label, as this should be probably documented somewhere at some point.

plextraktsync/plex_server.py
twolaw: yes i tested this

twolaw:
Thanks for your comments, I improved the code.
I think it is useless to mention this in readme file because average user will never have to go deaply into this configuration problem.
The error message is enough imho and now explicitly says : `Plex connection error: {}, local url {} didn't respond either. Check PLEX_LOCALURL in .env file.`

Maybe the error message should also talk about `plex_login` command ?

glensc (approved):

glensc:
Ok. I like the new version more (approved). But I can't test this myself.

twolaw:
tested successfully, I merge

keldian:
I'm noticing that when both BASEURL and LOCALURL are satisfied, PTS still listens on the BASEURL value. Would it not be more efficient to listen on the local host by default?

twolaw:
> Would it not be more efficient to listen on the local host by default?

Using localhost by default could not work with Plex instances requiring SSL and with shared libraries.

The PLEX_LOCALURL value is not populated by Plex but by the script. Therefore it is not 100% reliable.
The PLEX_BASEURL is provided by the Plex server itself and uses SSL which is much more reliable.

You can edit both PLEX_BASEURL and PLEX_LOCALURL as you wish in your .env file (exchange values if you prefer).

glensc:
What about saving all plex urls that plex-login is able to detect and sort them somehow? I think plex python library itself sorts them already by availability or speed

twolaw:
It can be a good idea to help users with dynamic ip address.

> sorts them already by availability or speed

and also local status and protocol (http vs https).

```xml
https://plex.tv/pms/resources.xml?includeHttps=1&X-Plex-Token=xxxxxxxxxxxxxxxxxxxxx

<Connection protocol="https" address="192.168.0.10" port="32400" uri="https://192-168-0-10.b25b53d5e99c40bc94126387dxxxxxxx.plex.direct:32400" local="1"/>
<Connection protocol="https" address="176.16.x.x" port="16422" uri="https://176-16-x-x.b25b53d5e99c40bc94126387dxxxxxxx.plex.direct:16422" local="0"/>
```

PR #850, opened 2022-03-27
Feature: Add support to override .env keys from actual environment

glensc:
Useful when the same `.env` file is used to initialize docker-compose but
want to change the `.env` file value at runtime.

This changes so that values of `.env` are always read even if the environment with the same name is already present.

```json
    "config": {
        "dotenv_override": true
    },
```

See https://github.com/Taxel/PlexTraktSync/discussions/849

### DIFF
--- a/plextraktsync/config.default.json
+++ b/plextraktsync/config.default.json
@@ -4,6 +4,9 @@
         "filename": "plextraktsync.log",
         "append": true
     },
+    "config": {
+        "dotenv_override": true
+    },
     "cache": {
         "path": "$PTS_CACHE_DIR/trakt_cache"
     },
--- a/plextraktsync/config.py
+++ b/plextraktsync/config.py
@@ -70,8 +70,9 @@ class Config(dict):
 
         config = self.load_json(self.config_file)
         self.merge(config, self)
+        override = self["config"]["dotenv_override"]
 
-        load_dotenv(self.env_file)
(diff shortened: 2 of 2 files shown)


glensc:
Maybe the default should be "on", so people who are aware of technical details can use this to let "optimize" not to re-read `.env` file with application if not needed.

twolaw:
Ok for default value to `true` but the key name `config` is not clear enough.
Maybe `env` or `dotenv` would be more explicit.

glensc:
52a11b76 & f12986f8, @twolaw how about now?

twolaw:
It looks good.

glensc:
Released in 0.18.23

glensc:
Doh. the code is broken:

```
$ plextraktsync info
PlexTraktSync Version: 0.18.23_main
Python Version: 3.10.0 (default, Nov 13 2021, 03:43:50) [GCC 10.2.1 20201203]
Plex API Version: 4.10.1
Trakt API Version: 3.4.0
Cache Dir: /app/config
Config Dir: /app/config
Log Dir: /app/config
Error: Error running info command: maximum recursion depth exceeded while decoding a JSON object from a unicode string
Error: Process completed with exit code 1.
```

glensc:
Fix: https://github.com/Taxel/PlexTraktSync/pull/856

twolaw:
You don't test before merge ?

PR #852, opened 2022-03-29
Feature: Add config.yml support

glensc:
Config load order:
- load config.defaults.yml
- if config.json present and config.yml is not:
    - load config.json
    - write config.yml with config.json contents
- if config.yml is missing, copy config.defaults.yml to it
- load config.yml present, load and merge it

x-ref:
- https://github.com/Taxel/PlexTraktSync/pull/850#issuecomment-1080383809
- https://python.land/data-processing/python-yaml


twolaw:
I suggest to remove the **config.json** file after copying to **config.yml** because users could keep editing config.json without understanding it's become useless (shortcut in text editor as an example).

glensc:
I'm against deleting. as that would not allow going back to older versions.

I could rename it to config.json.old with appropriate message.

twolaw:
> I could rename it to config.json.old with appropriate message.

Yes good idea, I approve this.

glensc:
> > I could rename it to config.json.old with appropriate message.
> 
> Yes good idea, I approve this.

9f5a9c43

twolaw (approved):

Issue #572, opened 2021-11-02
HDR metadata not being sent to Trakt collection


rfgamaral:
### The problem

I have lots of HDR movies in my collection, but the HDR information is not being set to Trakt for **any of them**.

Python is not really my thing, but I did a quick debugging and I believe the `hdr` function ([ref](https://github.com/Taxel/PlexTraktSync/blob/1063441838310229d1fee5061500a2069685f4c7/plex_trakt_sync/plex_api.py#L291)) is always returning `None` for all my movies.

### Expected behavior

HDR metadata should be sent with `plex_to_trakt.collection: true`.

### Install method

docker-compose

### Version

0.15.10

### Python Version

3.10.0

### Operating System and Version

GNU/Linux synology_apollolake_918+

glensc:
HDR is not my thing, but this is the PR that added the info extraction:
- https://github.com/Taxel/PlexTraktSync/pull/366

maybe there are details on how the extraction was designed.

you probably should come up with a solution yourself, add `pprint` to the function you referred to, and use `inspect` command to call for specific media

```py
    def hdr(self):
        """
        Set to dolby_vision, hdr10, hdr10_plus, or hlg
        """
        from print import pprint
        from sys import exit

        pprint(self.item)
        exit(33)

        try:
            stream = self.item.media[0].parts[0].streams[0]
```

```
./plextraktsync.sh inspect 1234
```

glensc:
@rfgamaral it's good info, that allows to investigate this deeper. since the property is only accessed when submitting item to collection, it's probably okay to fetch item fully. but I'm rather surprised why it wasn't autoloader by plexapi in first place.

rfgamaral:
I don't know what that means, but I hope you can fix it 😅

glensc:
Do you have any sample HDR video to provide? I tried searching with Search Engine but got just HDR images...

glensc:
I'm unable to reproduce with latest version, perhaps it was fixed with https://github.com/Taxel/PlexTraktSync/pull/652

In any case, of you have the id that has issues, first run with inspect:

```
./plextraktsync.sh inspect 16284
```

then modify source to print only on that id:

```diff
--- a/plextraktsync/media.py
+++ b/plextraktsync/media.py
@@ -138,6 +138,9 @@ def resolve_any(self, pm: PlexLibraryItem, tm=None):
         return None

     def resolve_guid(self, guid: PlexGuid, tm=None):
[...]

PR #870, opened 2022-04-03
Fix: Reload item when accessing parts for media

glensc:
Fixes https://github.com/Taxel/PlexTraktSync/issues/572

There's no major performance gain, as the parts are only accessed when the item is submitted to collection.

### DIFF
--- a/plextraktsync/plex_api.py
+++ b/plextraktsync/plex_api.py
@@ -227,7 +227,8 @@ class PlexLibraryItem:
 
     @property
     def parts(self) -> List[MediaPart]:
-        for media in self.item.media:
+        item = self.plex.fetch_item(self.item.ratingKey)
+        for media in item.item.media:
             yield from media.parts
 
     @flatten_list


M1kep:
What prevents this from being decorated as cached_property? In my mind if the first call to this always performs a refresh, the parts should then be static throughout the lifetime of that object

glensc:
please comment on the line (files changed view) or indicate by some other means of what line you have in mind.

and I don't understand what problem are you trying to solve there? the `parts()` method is called once for the lifetime of the sync for each item. the cache hit would be 0.

plextraktsync/plex_api.py
```diff
@@ -227,7 +227,8 @@ def collected_at(self):
 
     @property
     def parts(self) -> List[MediaPart]:
-        for media in self.item.media:
+        item = self.plex.fetch_item(self)
```
M1kep, line 230: Should this be `self.item.key`? 

Without grabbing the key I receive the below error:

```requests.exceptions.InvalidURL: Failed to parse:  https://<IP>.<ID>.plex.direct:32400<tmdb:509967:<Movie:129024:6-Underground>>```

M1kep: `self.plex` is also `None` when invoked via inspect. Not sure where that should be fixed, but locally was able to fix it by passing `plex=self` at:

- https://github.com/Taxel/PlexTraktSync/blob/51f22c1abffb886017b076887798a9d6c22ca529/plextraktsync/plex_api.py#L504

M1kep (commented):
It also looks like `self.ple

plextraktsync/plex_api.py
M1kep: It looks like #659 would fix the `None` issue?

glensc: 1. 2bca7ef3. 
2. `None` problem? Do you have an error?

M1kep: Can't repo at the moment as I'm running a full sync so it'll be about 40 minutes, but you should be able to reproduce by running `plextraktsync inspect <ID>`

glensc: Indeed, https://github.com/Taxel/PlexTraktSync/pull/659 is needed here

glensc: Rebased after https://github.com/Taxel/PlexTraktSync/pull/659  merge

PR #890, opened 2022-04-24
Add subcommand to install LaunchAgents wrapper for watch 

glensc:
Implement https://github.com/Taxel/PlexTraktSync/discussions/889 progammatically


plextraktsync/com.github.plextraktsync.watch.plist
```diff
@@ -0,0 +1,21 @@
...
+<dict>
+	<key>Label</key>
+	<string>com.github.plextraktsync.watch</string>
+	<key>ProgramArguments</key>
+	<array>
+		<string>plextraktsync</string>
+		<string>watch</string>
+	</array>
+	<key>LimitLoadToSessionType</key>
+	<string>Aqua</string>
```
glensc, line 13: why Aqua?

from this link `Background` seems more appropriate?
- https://stackoverflow.com/questions/9576287/what-are-the-os-x-session-types-and-what-do-they-mean

glensc:
the subcommand probably should have such call conventions:

```
# register/unregister
plextraktsync service register
plextraktsync service unregister

# to show the state
plextraktsync service show

# show usage
plextraktsync service
```

plextraktsync/com.github.plextraktsync.watch.plist
Bas-Man: I think I used aqua based on a old example. You are right that `Background` might be better. 

Setting a `nice` level might be a good idea though.

glensc: Send an example for nice level. perhaps using suggest code change feature:
- https://egghead.io/lessons/github-add-suggestions-in-a-github-pr-review

Bas-Man, line 13:
```suggestion
	<string>Background</string>
	<key>Nice</key>
	<integer>1</integer>
```

glensc: @Bas-Man I can't see the suggestion on the web, but had a glimpse of it in the email. Did you delete it?

Bas-Man:
@glensc the pending change request is on this pull request.

PR #907, opened 2022-05-05
Update README.md

kbgvirus:
added Docker-Compose crontab sync instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,14 @@ Installing from GitHub is considered developer mode and it's documented in
   $ crontab -e
   0 */2 * * * plextraktsync
   ```
+  Docker-Compose
 
+  Please change (container name) to your docker container name
+  
+  ```
+  $ crontab -e
+  0 */2 * * * sudo docker exec (container name) plextraktsync sync --sync all >/dev/null 2>&1
+  ```
 ## Sync settings
 
 To disable parts of the functionality of this software, look no further than


glensc:
this incomplete, where the `(container name)` would have appeared for it to be entered?

and the description is wrong, you are using docker not docker compose in your diffs.

for docker-compose to use in crontab you should put this line there, which is already documented:
- https://github.com/Taxel/PlexTraktSync#docker-compose

> 
> To run sync:
> 
> docker-compose run --rm plextraktsync

kbgvirus:
where (container name) individual needs to replace that with their container name.

docker exec executes the container docker-compose creates

`docker-compose run --rm plextraktsync` will stop the container after run, hence the use of my command entered

My instance is running correctly as intended, watch command stays running and sync runs every 30 minutes without killing the container.

not to worry I thought I would try help the community with similar use case

glensc:
> `docker-compose run --rm plextraktsync` will stop the container after run, hence the use of my command entered

unclear statement, that command will do the sync command exit, that's what you want to do.

You don't need to explain me things, i.e in merge request comments. I pointed out that merging this will create even more confusion because your example is incomplete, and there is already docker-compose command to do syncing in readme, you should use that command in your crontab. it's standalone without relying (not documented in your pull request) container creation.

also, fix PR title/body describing what are you trying to solve here.

PR #928, opened 2022-05-25
Show docker scheduler example as cron alternative

twolaw:
If users don't want or can't use **cron** to schedule syncs at set intervals, there is an alternative with docker : [Ofelia](https://github.com/mcuadros/ofelia/)

This PR adds in README.md an example of Ofelia docker scheduler to run a sync every 6h.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,30 @@ Installing from GitHub is considered developer mode and it's documented in
   0 */2 * * * plextraktsync
   ```
 
+* Instead of cron, a docker scheduler like [Ofelia](https://github.com/mcuadros/ofelia/) can also be used to run the script at set intervals.
+
+  A docker-compose example with a 6h interval:
+  
+  ```
+  services:
+    scheduler:
+      image: mcuadros/ofelia:latest
+      container_name: scheduler
+      depends_on:
+        - plextraktsync
+      command: daemon --docker
+      volumes:
+        - /var/run/docker.sock:/var/run/docker.sock:ro
+      labels:
+        ofelia.job-run.plextraktsync.schedule="@every 6h"
+        ofelia.job-run.plextraktsync.container="plextraktsync"
+    plextraktsync:
+      image: ghcr.io/taxel/plextraktsync:latest
+      container_name: plextraktsync
+      volumes:
+        - ./config:/app/config
+  ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -192,6 +192,32 @@ Installing from GitHub is considered developer mode and it's documented in
...
+    scheduler:
+      image: mcuadros/ofelia:latest
+      container_name: scheduler
+      depends_on:
+        - ptsync
+      command: daemon --docker
+      volumes:
+        - /var/run/docker.sock:/var/run/docker.sock:ro
+      labels:
+        ofelia.job-run.my_sync.schedule="@every 6h"
```
glensc, line 210: perhaps use "plextraktsync" instead of "exampleish" value "my_sync"

README.md
```diff
@@ -192,6 +192,32 @@ Installing from GitHub is considered developer mode and it's documented in
...
+        ofelia.job-run.my_sync.container="ptsync"
+    ptsync:
```
glensc, line 212: the main example uses "plextraktsync", use the same here for consistency.

twolaw: agreed

README.md
twolaw: it's to show the job name doesn't matter

glensc: not our job to show here that :)

very few will bother to edit and change that, most won't have ideas of implications. therefore better but some sane value here.

twolaw: ok I changed to plextraktync to make it simple

twolaw:
maybe TZ environment is overkill here too ?

README.md
```diff
@@ -192,6 +192,32 @@ Installing from GitHub is considered developer mode and it's documented in
...
+
+  A docker-compose example with a 6h interval:
+  
+  ```
+  services:
+    scheduler:
+      image: mcuadros/ofelia:latest
+      container_name: scheduler
+      depends_on:
+        - ptsync
```
glensc, line 205: needs to be changed too

glensc:
I think the TZ should be rather added to the other example too, first confirm it actually works, as maybe the image lacks tzdata package, so it has no effect.

twolaw:
Yes tzdata is missing in the official package. I remove TZ to not disturb average user.

Issue #888, opened 2022-04-23
"plextraktsync" does not work from crontab


KureOnGit:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists

### The problem

Hi,

I was running PlexTraktSync from my Download folder and a crontab service but decided to update for a more recent version. (zip)
That broke the manual run so I re-installed using pipx and it runs fine.
I have tried to add it to crontab however, and that does not seem to work.
Added a log and I get "Command not found", I imagine cron shell has no idea what it is I'm calling?

Should I add a path or something?
I have it running on Raspberry Pi OS.

Lines in cron I've tried
0 */12 * * * plextraktsync
0 */12 * * * plextraktsync >>/home/pi/Downloads/PlexTrakt.log 2>&1
0 */12 * * * /home/pi/Downloads/PlexTraktSync-main/plextraktsync.sh >>/home/pi/Downloads/PlexTrakt.log 2>&1  (old line)

### Expected behavior

Sync would run every 12h

### Steps to reproduce the behavior

1. Install plexsync
2. Try to have it run from crontab -e

### Install method

pipx (Recommended)

### Version

0.19.1

### Python Version

3.7.3

### Operating System and Version

Raspbian

glensc:
what's the output of `which plextraktsync` when in the terminal where you ran pipx install?

most likely the path of that script is not available from crontab, you can add that path to cron jobs:

```
PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin
```

or specify full path of your cronjob definition, ie the output of `which`.

KureOnGit:
Thank you very much for your reply.
Is that PATH a single path or that line adds multiple Paths?

The output of which was "/home/pi/.local/bin/plextraktsync" so I added that in.
I got a log created, so I think it worked. It has a bunch of errors that I don't think are related to running it from crontab. (seems to dislike hama, an agent that combines episodes and movies for anime libraries)
Will be checking in the next days if trakt gets updated in sync with what we see.
________________________________

KureOnGit:
Never really installed git, will give it a shot with desktop version.

PR #945, opened 2022-06-19
Update docs regarding cronjob setup

dennisvtang:
Additions were created based on conversation in #888.


twolaw:
Why don't you simply write your cronjob with full path :

`0 */2 * * * /home/pi/.local/bin/plextraktsync`

dennisvtang:
Yeah that would be way simpler. Do you think its enough to just change the existing example to what you suggested or should I include a note about using the `which` command to determine where the executable is?

twolaw:
You can add the note about `which` command.

dennisvtang:
I've simplified the documenation according to your suggestions. Please check again.

twolaw:
Ok good.
Can you replace `pi` with `user` (more generic) and squash commits.

twolaw:
Even better :
`0 */2 * * * $HOME/.local/bin/plextraktsync`
This is the standard pipx install destination of plextrakysync

glensc:
what a noisy PR. also, instead of the full path, you should just set $PATH to include the location of  the script:

```
PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:$HOME/.local/bin:/whatever/whenever

0 */2 * * * plextraktsync
```

Issue #956, opened 2022-06-28
Crash during sync : Read timed out. (read timeout=30)


radxrico:
correction below, hopefully

glensc:
Unclear what is the first trace from, seems you have truncated beginning of the output to see what it's from

But the second error is definitely because you have typed literal `ID` to the `inspect` command, you need to replace `ID` with the problematic id of the media, which may be the one seen in the sync script trace.

glensc:
You need to provide the output of the command run:
>  It gets to 95% done then it crashes

attach as .txt file if it's too long

radxrico:
Is this correct?

C:\Users\admin\Downloads\PlexTraktSync-0.20.1\PlexTraktSync-0.20.1>python -m plextraktsync inspect PlexLibrarySection:movie:Movies (Documentary)
PlexTraktSync [0.20.x]

Traceback (most recent call last):
  File "C:\Users\admin\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\requests\models.py", line 434, in prepare_url
    scheme, auth, host, port, path, query, fragment = parse_url(url)
  File "C:\Users\admin\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\urllib3\util\url.py", line 394, in parse_url
    return six.raise_from(LocationParseError(source_url), None)
  File "<string>", line 3, in raise_from
urllib3.exceptions.LocationParseError: Failed to parse: https://plex.pardesi.club:443PlexLibrarySection:movie:Movies

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.1520.0_x64__qbz5n2kfra8p0\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.1520.0_x64__qbz5n2kfra8p0\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Users\admin\Downloads\PlexTraktSync-0.20.1\PlexTraktSync-0.20.1\plextraktsync\__main__.py", line 18, in <module>
    cli()
[...]

PR #965, opened 2022-07-06
Feature: Add plex.timeout configuration

glensc:
Allow configuring plex timeout from this app config. for example:
- https://github.com/Taxel/PlexTraktSync/issues/956#issuecomment-1175871380


glensc:
@radxrico please don't hijack random issues/pull requests with your own topics. create your own issue. marked your comment as hidden.

PR #974, opened 2022-07-12
docs: add steps to configure the credentials in the docker container

Th3S4mur41:
Credentials cannot be configured when running the docker compose. The container has to be created with an interactive console first
https://github.com/Taxel/PlexTraktSync/discussions/970

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,17 +100,32 @@ or check output of [info command](#info-command).
 [appdirs]: https://pypi.org/project/appdirs
 [install-pipx]: https://github.com/pypa/pipx#install-pipx
 
-### Docker Compose
+### Docker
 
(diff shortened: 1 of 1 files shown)


Th3S4mur41:
> > Before running the container with Docker compose, you need to set it up with credentials
> 
> this is not true, you can run with docker-compose, no need to manually create the container.
> 
> ```
> docker-compose run --rm plextraktsync
> ```
> 
> this runs interactively (docker-compose default) and ask setup questions.
> 
Right, it didn't work in my case because I let the QNAP Container Station run my docker-compose which would kill the console before I had a chance to enter the email.
Running the docker compose in an SSH session worked as expected.
Thanks

PR #1030, opened 2022-08-30
Handle plex login BadRequest

twolaw:
Handle BadRequest instead of crashing

see #1028

### DIFF
--- a/plextraktsync/commands/plex_login.py
+++ b/plextraktsync/commands/plex_login.py
@@ -8,7 +8,7 @@ from typing import List
 import click
 from click import ClickException
 from InquirerPy import get_style, inquirer
-from plexapi.exceptions import NotFound, Unauthorized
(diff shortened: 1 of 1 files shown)


glensc:
@twolaw does it make sense to check for code 429 as well?

> plexapi.exceptions.BadRequest: (429) too_many_requests; https://plex.tv/users/sign_in.xml 

for example (right now) `Unauthorized` is a subclass of `BadRequest`:
- https://github.com/pkkid/python-plexapi/blob/bea32faa8b50d88499dda8602fc221d5e70910bd/plexapi/exceptions.py#L29

glensc:
seems the class has no `.code` property, the text in exception comes from plex.tv api

glensc (approved):

twolaw:
> @twolaw does it make sense to check for code 429 as well?

I think the user understands what to do after reading the error.
For docker users, the exit code 1 is enough I think.

Issue #1022, opened 2022-08-29
Error during Plex account access : Invalid token


kbgvirus:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists

### The problem

Hi Dev Team,

Since upgrading to the latest PlexTraktSync 0.21.1 I am getting the below error when trying to execute a sync. I have updated passwords due to Plex breach and have been working since. I have also tried re-authenticating with plex to get a new token

### Error trace / logs

```text
ERROR    Error during xxxxx account access: (422) unprocessable_entity; https://plex.tv/users/account <?xml version="1.0" encoding="UTF-8"?> <errors>   <error>Invalid token</error> </errors>
Sync Movie sections: [<PlexLibrarySection:movie:4K Movies>, <PlexLibrarySection:movie:Movies>]
Sync Show sections: [<PlexLibrarySection:show:TV Shows>]
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/app/plextraktsync/__main__.py", line 18, in <module>
    cli()
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/app/plextraktsync/cli.py", line 28, in wrap
    cmd(*args, **kwargs)
  File "/app/plextraktsync/commands/sync.py", line 60, in sync
    runner.sync(walker=w, dry_run=config.dry_run)
  File "/app/plextraktsync/sync.py", line 68, in sync
    self.plex_wl = {pm.guid: pm for pm in plex.watchlist()} or {}
[...]

PR #1031, opened 2022-08-30
Bypass if Plex watchlist is unavailable

twolaw:
Bypass watchlist sync if Plex watchlist is unavailable.

see #1022


glensc:
@twolaw why would the watchlist unavailable? older plex server?

twolaw:
Like in #1022 you can have correct access to your Plex server but no access to your Plex account.

glensc:
yes, you linked to https://github.com/Taxel/PlexTraktSync/issues/1022 already, but it doesn't explain there why they don't have access.

PR #1038, opened 2022-09-02
Add plex account token for shared resource

twolaw:
Account token ≠ Resource access token for users with shared library

This PR adds the **plex_account_token** to `.env` file for Plex account users using a shared resource.


glensc:
why would you need account token? can you describe what process or code needs it?

twolaw:
If you read the code, you can see it's used to get the watchlist.

glensc:
could you just put that info to PR body please

PR #1040, opened 2022-09-04
Feature: Conditional library walk

twolaw:
Do not walk across Plex libraries if it's not needed


glensc:
wouldn't it be simpler to skip calling `runner.sync` if nothing to walk? i.e then wouldn't need to add the condition in Sync class and in sync command. i.e check the state and return early in command/sync.py

twolaw:
Nothing to walk but something to sync 😉

glensc:
> Nothing to walk but something to sync 😉

please explain.

twolaw:
#1019

Issue #966, opened 2022-07-06
Feature request: Progress Reset


EffeF:
### Feature description
This VIP feature allows you to reset a show's progress and track rewatches from your progress page and dashboard up next.
I expect progress to be reset on Plex as well, so that rewatching can be done easily without having to go searching for the episode within a series already watched.

### References
https://trakt.tv/vip/progress-reset
https://support.trakt.tv/support/solutions/articles/70000628748-how-to-reset-watched-progress
https://trakt.docs.apiary.io/#reference/shows/watched-progress/get-show-watched-progress

EffeF:
> Yes.

Then manually marking it with unwatched is not the right solution, at least to my way of use.

> PS: I don't know if we want to add code designed for trakt VIP members only.

I don't know how many VIP users use this script, surely there are some.

Trakt's API provides the user's VIP status, so in this respect just do a check and disable this and possibly other VIP features.

At this point whether to implement it or not just depends on the time required to implement it and whether there are other VIP users interested...

twolaw:
> Then manually marking it with unwatched is not the right solution, at least to my way of use.

We can add a condition reading show's **reset_at** value before matching watched status of episodes.
A reset show must not re-mark as watched episodes seen before the reset.

EffeF:
This could be an interesting solution, more than the smart playlist in my opinion.

Therefore, the process would be:
1. reset the progress on Trakt
2. mark as unwatched on Plex
3. if `reset_at` > `last_watched_at` PlexTraktSync not re-mark as watched episodes

Ideally, step 2 would be automated as well, but even so it is a great solution for managing rewatches

EffeF:
>  You loose Plex view history, but it's still available on trakt.

This is not a problem for me. In my use Plex is just a media server, my history is on Trakt.

> So they know what they did on purpose.

Agreed. Maybe as an option disabled by default in the configuration file? So it would be something that just happens only as a choice on purpose.

Hubro:
> About step 2, I'm not fond of automatically mark all episodes unwatched because users will loose their Plex view history after a trakt reset

This seems like the expected outcome to me. If a user resets a show in Trakt, the intuitive result is that it should become unwatched in Plex.

It also seems to me that if somebody uses Trakt, it's unlikely that they care very much about the long term local state in Plex. If I want to check how many times I've watched episode, I will check Trakt, not Plex.

If you think there are users that would be bothered by this (I don't) then you could hide this behind an option.

Something like:

[...]

PR #1046, opened 2022-09-06
Docker instructions for PR install

glensc:
Taken from https://github.com/Taxel/PlexTraktSync/issues/966#issuecomment-1236524825


glensc:
There are tons of variations how to do it in docker, so I've included one that creates temporary container. temporary in sense if you exit, the installation will be lost, so it doesn't affect existing installations in any way.

twolaw (approved):

PR #1084, opened 2022-10-02
Feature: Add progressbar to watch status

glensc:


twolaw:
I suggest to use `duration` instead of `length` word.

image

glensc:
the field is not even needed for the progressbar. i wanted to make a progressbar which shows current time and max time, but did not find existing example for that and implementing on my own seemed complicated. i thought it would be needed for progressbar. but right now it's implemented as automatic, that progressbar itself estimates the time. 

so I'll extract it from this pr. I wanted to have a different (but compact) name in case want to access original duration without the `.item` subkey.

PR #1086, opened 2022-10-02
Refactor PlexServerConnection to support any number of server urls

glensc:
A preparation of storing multiple server urls in a separate servers.yml config.

### DIFF
--- a/plextraktsync/plex_server.py
+++ b/plextraktsync/plex_server.py
@@ -1,92 +1,94 @@
 from functools import partial
+from typing import List
 
 import plexapi.server
 
 from plextraktsync.config import PLEX_PLATFORM
 from plextraktsync.decorators.nocache import nocache
-from plextraktsync.factory import factory
+from plextraktsync.factory import Factory
 from plextraktsync.logging import logger
 
 
 class PlexServerConnection:
-    @staticmethod
+    def __init__(self, factory: Factory):
+        self.factory = factory
+
+    @property
+    def timeout(self):
+        return self.factory.config()["plex"]["timeout"]
+
+    @property
+    def session(self):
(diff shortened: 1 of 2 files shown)


plextraktsync/plex_server.py
```diff
@@ -1,92 +1,110 @@
...
+                #   CertificateError(
+                #     "hostname '127.0.0.1' doesn't match '*.5125cc430e5f1919c27226507eab90df.plex.direct'"
+                #    )
+                #  )
+                # )
+                if "doesn't match '*." in message and ".plex.direct" in url:
+                    url = self.extract_plex_direct(url, message)
+                    logger.warning(f"Trying with url: {url}")
+                    urls.append(url)
+                    self.save_new_url(url, urls[-1])
```
glensc, line 61: I don't like this saving over the new config here. doesn't really fit with what the method is doing here. I think it's fine to do that every time the error is thrown.

glensc:
Seems to be working most of the time. Tested common scenarios like timeout and connection refused and invalid cert. Of course, can fix in follow-up pull requests if some unhandled cases happen.

For example, I don't really understand what the use case was for .plex.direct cert error. seems like updating only the hash portion of .plex.direct URL. but how that can happen? usually, if it changes then the IP part changes too, like users manually try to invent a plex URL and we need to handle such in this project?

more functional example would be to turn any ip to .plex.direct direct url, to have https url, i.e including encoding the ip and the hash, but such code is not here.

probably there's the answer in old pull requests or issues, but I'm not invested enough to dig it out.

also, i would prefer if such super smart code is in some specific command, dedicated to it.  not in default connection method. altho why not just run plex-login and be done with it?

PR #1087, opened 2022-10-02
Feature: Add multiple servers support

glensc:
Allow easy switching between multiple Plex servers.

This moves servers definition to `servers.yml` file:

```yml
servers:
  default:
    token: ~
    urls:
      - http://localhost:32400
```

in `.env`:
```sh
# Name of active server from servers.yml
PLEX_SERVER=default
```

`urls` can be an unlimited amount of urls to try.

`plex-login` command can be used to add new servers, it will add/update servers.yml and update `PLEX_SERVER` in `.env`.

Future ideas:
- allow specify server via commandline by name
- walk multiple servers
- per server config overrides

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 /docker-compose.override.yml
 /last_update.log
 /plextraktsync.log
+/servers.yml
 /tests/.env
 /tests/config.json
 /trakt_cache.sqlite
--- a/plextraktsync/commands/plex_login.py
+++ b/plextraktsync/commands/plex_login.py
@@ -10,6 +10,8 @@ from plexapi.exceptions import BadRequest, NotFound, Unauthorized
 from plexapi.myplex import MyPlexAccount, MyPlexResource, ResourceConnection
 from plexapi.server import PlexServer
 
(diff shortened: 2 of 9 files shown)


glensc:
I'm just wondering are `PLEX_USERNAME`, `PLEX_OWNER_TOKEN`, `PLEX_ACCOUNT_TOKEN` part of server selection or not?

minimum (and easiest to implement) needed to switch servers is `urls` and `token`:
```yml
servers:
  default:
    token: ~
    urls:
      - http://localhost:32400
```

I personally have one plex account but multiple servers.

twolaw:
Needs to be tested for :
- standard Plex user
- home user
- user with shared library

glensc:
> Needs to be tested for :
> 
> - standard Plex user
> - home user
> - user with shared library

why? nothing changes in the authentication, still logging in with url + token

smailpouri:
Can shared servers be added to track when watching from a shared server?
When going through the process of plextraktsync plex-login to add the shared server for a second time, it automatically stops tracking from my own. instead only tracks from the shared server.

PR #1126, opened 2022-10-23
Fix: Force logging.basicConfig to logging to work in pytest

glensc:
Extracted from https://github.com/Taxel/PlexTraktSync/pull/716

### DIFF
--- a/plextraktsync/logging.py
+++ b/plextraktsync/logging.py
@@ -26,7 +26,7 @@ def initialize(config):
         file_handler,
         console_handler,
     ]
-    logging.basicConfig(handlers=handlers, level=log_level)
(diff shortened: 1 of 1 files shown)


glensc:
Hmm, the `force` seems to be missing in python 3.7

```
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/logging/__init__.py", line 1919, in basicConfig
    raise ValueError('Unrecognised argument(s): %s' % keys)
```
- https://github.com/Taxel/PlexTraktSync/actions/runs/3307539321/jobs/5459250021#step:6:34

but I don't see the version note in docs:
- https://docs.python.org/3/library/logging.html#logging.basicConfig

PR #1127, opened 2022-10-23
Update to pytrakt 3.4.5

glensc:
Update to use package from PyPI:
- https://pypi.org/project/pytrakt/

Because the source version still ended up with an error:
- https://github.com/s0undt3ch/setuptools-declarative-requirements/issues/8#issuecomment-1288145464

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -15,6 +15,7 @@ oauthlib = "==3.2.2"
 plexapi = "==4.13.0"
 python-dotenv = "==0.21.0"
 python-git-info = "==0.8.0"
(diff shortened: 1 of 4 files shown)


glensc:
@twolaw create yourself an account at [pypi](https://pypi.org/) and inform me about your handle, so I can add you as Owner for the package in PyPI. I asked you the same to add you to PlexTraktSync but never got reply from you.

twolaw:
I just created my account as `twolaw` at pypi.org
You can add me, thanks

PR #1136, opened 2022-10-24
Refactor: Add show property to PlexLibraryItem

glensc:
It's only valid if the type is episode.


glensc:
Actually, need Trakt object for Lookup class:
```py
lookup = TraktLookup(self.trakt)
```

might have analyzed wrong.

twolaw:
Where is this `show` property needed ?

glensc:
> Where is this show property needed ?

I already answered, it's need for episode object to create TraktLookup object.

glensc:
Also, seems PlexApi Episode object has `show()` method:
- https://github.com/pkkid/python-plexapi/blob/4.13.0/plexapi/video.py#L926-L928

altho it's not efficient one, it's always fetching from plex server.

PR #1141, opened 2022-10-25
Feature: Add --no-cache flag to inspect command

glensc:
Could be useful for debugging.


glensc:
This can likely be created very simple requests cache with rules that disable all caching after https://github.com/Taxel/PlexTraktSync/pull/1221

glensc:
Replaced with https://github.com/Taxel/PlexTraktSync/pull/1287

Issue #1157, opened 2022-10-31
Something broken in latest version.


twhiting9275:
Trying to get this working on Windows 10 (python 3.8). Unfortunately, no dice. Downloaded from Github, and latest version

```
D:\pts>C:\Python38\python.exe -m plextraktsync
Traceback (most recent call last):
  File "C:\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "D:\pts\plextraktsync\__main__.py", line 18, in <module>
    cli()
  File "C:\Python38\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "C:\Python38\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "C:\Python38\lib\site-packages\click\core.py", line 1635, in invoke
    rv = super().invoke(ctx)
  File "C:\Python38\lib\site-packages\click\core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Python38\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "C:\Python38\lib\site-packages\click\decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "D:\pts\plextraktsync\cli.py", line 53, in cli
    sync()
  File "C:\Python38\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "C:\Python38\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "C:\Python38\lib\site-packages\click\core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Python38\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "D:\pts\plextraktsync\cli.py", line 24, in wrap
    module = importlib.import_module(f".commands.{name}", package=__package__)
  File "C:\Python38\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
[...]

Issue #1154, opened 2022-10-29
ImportError: cannot import name 'BadResponseException' from 'trakt.errors'


solid-pixel:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists

### The problem

Hello, thanks for pushing so many updates, you guys are doing a great job.

It seems like every 0.23.x version is throwing this error for me whenever I run `plextraktsync.sh`. 
0.22.4 is the last one to work flawlessly.

Also I'm not sure if it's the expected behavior, but running `plextraktsync.sh --version` shows `0.23.0dev0` instead of `0.23.7`:

```console
mrrobot@polaris:/home1/mrrobot/custom/PlexTraktSync$ git describe --tags
0.23.7
mrrobot@polaris:/home1/mrrobot/custom/PlexTraktSync$ ./plextraktsync.sh --version
PlexTraktSync 0.23.0dev0: e7ec3755: Bring back media.show property mistakenly removed in 1eaf574 (#1152) @2022-10-29 12:19:38
```

Running `plextraktsync.sh info` or any other argument also throws the error showed below.

### Error trace / logs

```console
mrrobot@polaris:/home1/mrrobot/custom/PlexTraktSync$ ./plextraktsync.sh
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home1/mrrobot/custom/PlexTraktSync/plextraktsync/__main__.py", line 18, in <module>
    cli()
  File "/home/mrrobot/.local/lib/python3.7/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/mrrobot/.local/lib/python3.7/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/mrrobot/.local/lib/python3.7/site-packages/click/core.py", line 1635, in invoke
    rv = super().invoke(ctx)
  File "/home/mrrobot/.local/lib/python3.7/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mrrobot/.local/lib/python3.7/site-packages/click/core.py", line 760, in invoke
[...]

Issue #1144, opened 2022-10-27
Mark_Watched: Unsupported media type: episodes error on sync


cjsage52:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists

### The problem

I get the following error when syncing: "Error: Error running sync command: mark_watched: Unsupported media type: episodes"

### Error trace / logs

```text
INFO     Adding to collection: <tmdb:770610:Episode:1154067:Cinderella-Man-s01e01>
INFO     Adding to collection: <tmdb:770611:Episode:1154068:Cinderella-Man-s01e02>
INFO     Adding to collection: <tmdb:770612:Episode:1154069:Cinderella-Man-s01e03>
INFO     Adding to collection: <tmdb:770613:Episode:1154070:Cinderella-Man-s01e04>
INFO     Adding to collection: <tmdb:770614:Episode:1154071:Cinderella-Man-s01e05>
INFO     Adding to collection: <tmdb:770615:Episode:1154072:Cinderella-Man-s01e06>
INFO     Adding to collection: <tmdb:770616:Episode:1154073:Cinderella-Man-s01e07>
INFO     Adding to collection: <tmdb:770617:Episode:1154074:Cinderella-Man-s01e08>
INFO     Adding to collection: <tmdb:770618:Episode:1154075:Cinderella-Man-s01e09>
INFO     Adding to collection: <tmdb:770619:Episode:1154076:Cinderella-Man-s01e10>
INFO     Adding to collection: <tmdb:770620:Episode:1154077:Cinderella-Man-s01e11>
INFO     Adding to collection: <tmdb:770621:Episode:1154078:Cinderella-Man-s01e12>
INFO     Adding to collection: <tmdb:770622:Episode:1154079:Cinderella-Man-s01e13>
INFO     Adding to collection: <tmdb:770623:Episode:1154080:Cinderella-Man-s01e14>
INFO     Adding to collection: <tmdb:770624:Episode:1154081:Cinderella-Man-s01e15>
INFO     Adding to collection: <tmdb:770625:Episode:1154082:Cinderella-Man-s01e16>
INFO     Adding to collection: <tmdb:1338967:Episode:1127207:Love-Sick:-The-Serie-s01e01>
INFO     Marking as watched in Trakt: <tmdb:1338967:Episode:1127207:Love-Sick:-The-Serie-s01e01>
[...]

PR #1158, opened 2022-10-31
Docs: Make clearer note about unsupported install methods

glensc:
There has been an increasing volume of people making reports of installation failure, while they have not fully followed the installation documentation.

This makes it clear that only pipx and docker image are supported. other install methods are just documented as a convience, but without support.

- https://github.com/Taxel/PlexTraktSync/issues/1157
- https://github.com/Taxel/PlexTraktSync/issues/1154
- https://github.com/Taxel/PlexTraktSync/issues/1144

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -81,11 +81,11 @@ body:
       label: Install method
       description: How did you install PlexTraktSync
       options:
-        - pipx (Recommended)
-        - pip
+        - pipx
         - docker-compose
-        - git clone
-        - zip download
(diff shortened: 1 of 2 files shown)


bulletproof2k:
That's not so. The installation method for Windows worked in one of the versions. I may be wrong, but most likely one of these versions is 0.22.4 or 0.23.4.

glensc:
I can not support that windows installation step, windows install was contributed by a user, same for unraid. install with pipx and be done with the wrong/missing/conflicting dependency problems.

twolaw (commented):
Looks good. Also you can add `(not supported)` to items in bug.yml template install list.

Issue #1161, opened 2022-11-01
Trakt official lists raise : 'TraktList' object has no attribute 'trakt_items'


bulletproof2k:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists

### The problem

```
Traceback (most recent call last):
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\IDE\.py\PlexTraktSync-0.23.4\plextraktsync\__main__.py", line 18, in <module>
    cli()
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1635, in invoke
    rv = super().invoke(ctx)
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\site-packages\click\decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "D:\IDE\.py\PlexTraktSync-0.23.4\plextraktsync\cli.py", line 53, in cli
    sync()
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\victo\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
[...]

PR #1162, opened 2022-11-02
Add trakt official lists support

twolaw:
With this PR, the script handles trakt lists with their id instead of there creator name.
This allows to use user lists and official lists with the same function.
Official lists are not identical to user lists because they are not created bu users and cannot be called with `/users/(user)/lists/(list)`

Ideally, the pytrakt module would add official lists support.

### DIFF
--- a/plextraktsync/sync.py
+++ b/plextraktsync/sync.py
@@ -131,7 +131,7 @@ class Sync:
 
         if self.config.sync_liked_lists:
             for lst in self.trakt.liked_lists:
-                listutil.addList(lst["username"], lst["listname"])
+                listutil.addList(lst["listid"], lst["listname"])
 
         if self.config.need_library_walk:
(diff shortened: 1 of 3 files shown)


glensc:
install from pull request instructions:
- https://github.com/Taxel/PlexTraktSync/blob/143ab6c664040d9ace4e9b671a13feefc1ae68c8/CONTRIBUTING.md#install-code-from-pull-request

glensc:
I asked @bulletproof2k to test this pr:
- https://github.com/Taxel/PlexTraktSync/issues/1161#issuecomment-1301908510
- https://github.com/Taxel/PlexTraktSync/issues/1161#issuecomment-1301915009
- https://github.com/Taxel/PlexTraktSync/issues/1161#issuecomment-1301923495

Issue #124, opened 2021-03-11
Avoid delete/create playlists


glensc:
Even I think the original problem is that Plex does not allow Playlist modification, so this repo implemented delete+create, but this can be optimized, if the playlist is not to be updated, skip its modification.

Looks like this playlist id increments with each run. Based on symptoms, seems the sync script creates new playlist each time it runs:

```
-WARNING:Completed section sync in 20.8 seconds
-WARNING:BadRequest (204) no_content http://acai.local:32400/playlists/10170;
+WARNING:Completed section sync in 22.3 seconds
+WARNING:BadRequest (204) no_content http://acai.local:32400/playlists/10171;
```

_Originally posted by @glensc in https://github.com/Taxel/PlexTraktSync/issues/112#issuecomment-791972992_

glensc:
There are methods that support editing:
- https://python-plexapi.readthedocs.io/en/latest/modules/playlist.html#plexapi.playlist.Playlist.addItems
- https://python-plexapi.readthedocs.io/en/latest/modules/playlist.html#plexapi.playlist.Playlist.removeItems
- https://python-plexapi.readthedocs.io/en/latest/modules/playlist.html#plexapi.playlist.Playlist.moveItem

so, if someone comes up with the shortest list of actions to keep two lists synchronized, then it could be done. I guess the original author just was lazy and wen't with delete+create option.

At a minimum, could at least skip deleting if the list is identical (no changes since the last sync).

PR #1183, opened 2022-11-09
Refactor: Update Plex playlists not delete+create

glensc:
Fixes https://github.com/Taxel/PlexTraktSync/issues/124

This allows you edit Plex lists, add them description:
- https://github.com/pkkid/python-plexapi/issues/1054

and it will be preserved as long as you keep the name intact.

Playlist contents is still cleared with each update to supply order by PTS.
it would be overly complicated to do compare and run only three available methods:
- addItems
- deleteItems
- reorderItems

besides, reorder items can take only one item a time, so lots of api calls

### DIFF
--- a/plextraktsync/plex_api.py
+++ b/plextraktsync/plex_api.py
@@ -10,6 +10,7 @@ from plexapi.exceptions import BadRequest, NotFound, Unauthorized
 from plexapi.library import LibrarySection
 from plexapi.media import AudioStream, MediaPart, SubtitleStream, VideoStream
 from plexapi.myplex import MyPlexAccount
+from plexapi.playlist import Playlist
 from plexapi.server import PlexServer, SystemAccount, SystemDevice
 from plexapi.video import Episode, Movie, Show
 from trakt.utils import timestamp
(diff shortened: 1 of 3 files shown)


plextraktsync/plex_api.py
```diff
@@ -607,16 +608,17 @@ def rate(self, m, rating):
...
+        """
+        Updates playlist (creates if name missing) replacing contents with items[]
+        """
         try:
-            self.plex.playlist(name).delete()
-        except (NotFound, BadRequest):
-            logger.debug(f"Playlist '{name}' not found, so it could not be deleted")
+            playlist: Playlist = self.plex.playlist(name)
+        except NotFound:
+            playlist = self.plex.createPlaylist(name)
```
NGDM, line 618: The createPlaylist method requires items to be passed.
Right now it's giving a BadRequest error when creating a new playlist.
I can check to make a PR for this as well.

glensc: oh ok, sorry my bad for not testing that use case, does it accept an empty list `[]`?

luckily this code is not yet released (no tag yet), so no harm done.

NGDM: > oh ok, sorry my bad for not testing that use case, does it accept an empty list `[]`?

It does not apparently. That sucks, because if a user likes a list for which he has no mutual items, the script will crash.
We'll need to add a check if the list has items. 
I'll make a new PR for this.

PR #1185, opened 2022-11-09
Allow liked lists to sync without watchlist sync

NGDM:
PR for Feature request: Allow syncing of Liked Lists without needing Watchlist sync.

Resolves #1182

PR #1191, opened 2022-11-12
Fix watchlist update logic

twolaw:
Issue : if you wanted to sync both `liked lists` and `watchlist`, the watchlist would never sync (because of a `else` logic).
This PR fix this issue.

Also uses only one log line for each playlist update.

Tested successfully.


glensc:
I guess this is improvement to https://github.com/Taxel/PlexTraktSync/pull/1185

PR #1195, opened 2022-11-14
Feature: Add support for description for update_playlist

glensc:
Now Playlists created in Plex will have same description as in Trakt


glensc:
Could additionally add some more meta to the description, like:
- last_updated field
- url of the trakt list
- user, creator
- sort by info

I'm leaving these out of this PR for now.

PR #1217, opened 2022-11-23
Feature: Save all urls to config from plex-login

glensc:
Uses [`preferred_connections`](https://github.com/pkkid/python-plexapi/blob/3d3f9da5012428f5d703cf9f8e95f6aa10673ea6/plexapi/myplex.py#L1309-L1314) which describes as:
> Returns a sorted list of the available connection addresses for this resource.
> Often times there is more than one address specified for a server or client.
> Default behavior will prioritize local connections before remote or relay and HTTPS before HTTP.
> 

Writes all urls in order like:

```yml
servers:
  default:
    urls:
    - https://192-168-0-1.5eaf640a56769cf3a72fd93ac9711e40.plex.direct:32400
    - http://192.168.0.1:32400
    - https://8-8-8-8.5eaf640a56769cf3a72fd93ac9711e40.plex.direct:12521
    - http://8.8.8.8:12521
    - http://localhost:32400
```

### DIFF
--- a/plextraktsync/commands/plex_login.py
+++ b/plextraktsync/commands/plex_login.py
@@ -46,6 +46,7 @@ def server_urls(server: MyPlexResource):
     """
 
     yield from [c.uri for c in server.connections]
+    yield from [c.httpuri for c in server.connections]
     yield local_url()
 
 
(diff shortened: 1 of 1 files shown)


glensc:
not sure if the `local_url()` is even needed anymore.

PR #1220, opened 2022-11-23
Feature: Add http caching to methods

glensc:
Rather using `@nocache` decorator, specify with `@http_cache` for how long to cache.


glensc:
But this is a wrong approach. if some method makes two requests, you may still get unwanted behavior.

Should use url patterns instead:
- https://requests-cache.readthedocs.io/en/stable/examples.html#url-patterns

glensc:
Replaces https://github.com/Taxel/PlexTraktSync/pull/1221

Issue #1060, opened 2022-09-14
Numerous Duplicate marking as Played on Plex Media Server


gbooker:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists

### The problem

Plex developer here:

In diagnosing a problem with Plex Media Server and the View State Sync service on plex.tv, we have noticed that several users of PlexTraktSync are having issues where several movies are being marked as played either every 1 or 2 hour.  Other entries in the PMS logs indicate that this is the frequency of which PlexTraktSync is run against the server.

Note: I do not know the user's installation methods/OS/version as I am looking at the media server's logs when I noticed the problem.

### Error trace / logs

```text
From the PMS logs (IP/Username/Servername obfuscated):

Sep 14, 2022 17:00:10.376 [0x7fc7cbc82b00] DEBUG - Request: [IP:Port (Subnet)] GET /:/scrobble?key=48038&identifier=com.plexapp.plugins.library (5 live) #d81d2 TLS GZIP Signed-in Token (Username) (Server)
Sep 14, 2022 17:00:10.377 [0x7fc7cbc82b00] DEBUG - [Req#d81d2] Library item 48038 'The Tinder Swindler' got played by account 1!
Sep 14, 2022 17:00:10.377 [0x7fc7d150db00] DEBUG - Completed: [IP:Port] 200 GET /:/scrobble?key=48038&identifier=com.plexapp.plugins.library (5 live) TLS GZIP 1ms 195 bytes (pipelined: 11)
```
Same request occurs every hour as far back as the logs go.

### Expected behavior

PlexTraktSync should not record an items as played which is already played.

### Steps to reproduce the behavior

Unsure of how this circumstance is occurring but we have several users who are using PlexTraktSync which are having the same set of movies repeatedly being marked as played.

### Workarounds

We are telling user to try disabling PlexTraktSync and the items being repeatedly being marked as played ceases.

### Install method

pipx (Recommended)

### Version

Unknown

### Python Version

Unknown

### Operating System and Version

Unknown

twolaw:
Can we get more info from the users :

- version of the script
- which set of movies (why not all)
- a debug trace of a problematic sync
- an `inspect` command run on a problematic movie
- test of relogin and clear cache
- test of downgrade script

gbooker:
I don't have the info from users but we did ask them if they could provide info to this GHI.
[...]

PR #1221, opened 2022-11-23
Refactor: Setup url cache for trakt.tv by patterns

glensc:
Replaces https://github.com/Taxel/PlexTraktSync/pull/1220

Removes `@nocache` annotation and moves them to be setup via rules:

```yml
##### Advanced settings below this line, don't edit unless you know what you're doing #####
http_cache:
  # Value "0" means never cache, value "-1" means never expires
  # anything else means how long to cache
  #
  # NOTE: If there is more than one match, the first match will be used in the order they are defined
  policy:
    "*.trakt.tv/shows/*/seasons": 0
    "*.trakt.tv/sync/collection/shows": 0
    "*.trakt.tv/sync/watched/shows": 0
    "*.trakt.tv/users/*/collection/movies": 0
    "*.trakt.tv/users/*/collection/shows": 0
    "*.trakt.tv/users/*/ratings/*": 0
    "*.trakt.tv/users/*/watched/movies": 0
    "*.trakt.tv/users/*/watchlist/movies": 0
    "*.trakt.tv/users/*/watchlist/shows": 0
    "*.trakt.tv/users/likes/lists": 0
    "*.trakt.tv/users/me": 0
```

This allows to setup rule to cache some watchlist forever.

In the future could add default cache key (removed from this branch for now):

```yml
http_cache:
  # By default, cached responses expire in an hour
  default_expire_after: 1 hour
```

### DIFF
--- a/plextraktsync/commands/config.py
+++ b/plextraktsync/commands/config.py
@@ -7,3 +7,6 @@ def config(print=print):
 
     print(f"# Config File: {config.config_yml}")
     config.dump(print=print)
+
+    print("# HTTP Cache")
+    config.http_cache.dump(print=print)
--- a/plextraktsync/config.default.yml
+++ b/plextraktsync/config.default.yml
@@ -61,14 +61,14 @@ xbmc-providers:
   shows: tvdb
 
 ##### Advanced settings below this line, don't edit unless you know what you're doing #####
-http_cache:
+#http_cache:
   # Value "0" means never cache, value "-1" means never expires
   # anything else means how long to cache
   #
   # https://requests-cache.readthedocs.io/en/stable/user_guide/expiration.html#url-patterns
   #
   # NOTE: If there is more than one match, the first match will be used in the order they are defined
-  policy:
+#  policy:
 #    "*.trakt.tv/users/me": 1 day
 #    "*.trakt.tv/users/likes/lists": 0
 
--- a/plextraktsync/config.default.yml
+++ b/plextraktsync/config.default.yml
@@ -1,21 +1,6 @@
 cache:
   path: $PTS_CACHE_DIR/trakt_cache
 
-http_cache:
-  never_cache:
-    # Requests matching these patterns will not be cached
-    - "*.trakt.tv/shows/*/seasons"
-    - "*.trakt.tv/sync/collection/shows"
-    - "*.trakt.tv/sync/watched/shows"
-    - "*.trakt.tv/users/*/collection/movies"
-    - "*.trakt.tv/users/*/collection/shows"
-    - "*.trakt.tv/users/*/ratings/*"
-    - "*.trakt.tv/users/*/watched/movies"
-    - "*.trakt.tv/users/*/watchlist/movies"
-    - "*.trakt.tv/users/*/watchlist/shows"
-    - "*.trakt.tv/users/likes/lists"
-    - "*.trakt.tv/users/me"
-
 excluded-libraries:
   - Private
   - Family Holidays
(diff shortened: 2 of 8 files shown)


glensc:
This could even theoretically speed up sync, because `@nocache` is not called. but now `requests-cache` has to walk the `urls_expire_after` dict.

glensc:
I think `@nocache` from methods that do `@post` can be just dopped, request-cache doesn't cache `POST` requests.

glensc:
I put the rules to `config.default.yml`, so users can add their own rules. the current implementation is that if the list is merged then the whole contents are overwritten. should the lists be concatenated instead?

i.e put to your `config.yml`:
```yml
http_cache:
  never_cache:
    - "*.plex.direct:*/library/metadata/7602/allLeaves"
```

and with `plextraktsync config` you will see that the defaults are overwritten, so you would need to copy items from default config as well.

therefore I think maybe need to merge values if type is "list"?

glensc:
Uh, that would be a problem if they really copy to their config, or it's copied automatically on first install. as when we publish updates to the default config, it won't apply because user configuration overrides.

glensc:
Maybe should change the structure to a `dict`:

```yml
http_cache:
  # Value "0" means never cache, value "-1" means never expires
  # anything else means how long to cache
  #
  # NOTE: If there is more than one match, the first match will be used in the order they are defined
  policy:
    "*.trakt.tv/shows/*/seasons": 0
    "*.trakt.tv/sync/collection/shows": 0
    "*.trakt.tv/sync/watched/shows": 0
    "*.trakt.tv/users/*/collection/movies": 0
    "*.trakt.tv/users/*/collection/shows": 0
    "*.trakt.tv/users/*/ratings/*": 0
    "*.trakt.tv/users/*/watched/movies": 0
    "*.trakt.tv/users/*/watchlist/movies": 0
    "*.trakt.tv/users/*/watchlist/shows": 0
    "*.trakt.tv/users/likes/lists": 0
    "*.trakt.tv/users/me": 0

    # Plex URLs can be cached too:
    "*.plex.direct:*/library/metadata/*/allLeaves": 1 day
    "*/library/metadata/*/allLeaves": 30 days
```

twolaw:
Yes, good idea.
What about urls not in this dict?

glensc:
if you don't specify any value for URL pattern, then it's cached forever. unless we say default_expire value. then default_expire would be the expire time.

glensc:
So where did we end up here? Perhaps move the config to code for now?

glensc:
Added `default_policy` to code, disabled entries from default config.

twolaw:
Maybe add a plex url as example too in `config.default.yml`.
This would help users wanting to refresh metadata after a Plex [Fix match](https://support.plex.tv/articles/201018497-fix-match-match/).

https://github.com/Taxel/PlexTraktSync/issues/1060#issuecomment-1248435608

glensc:
Expire flag is more useful for removing an cache entry:
- https://github.com/Taxel/PlexTraktSync/pull/1277

glensc:
also, i don't think Fix Match is cached, as processing sync will do `/library/sections/1/all?X-Plex-Container-Start=0&X-Plex-Container-Size=100` calls which is not cached.

glensc:
and this PR affects only trakt requests.

glensc:
and plex URLs are problematic, they don't always match `*.plex.direct:*`, you could have IP or even a totally custom DNS name. therefore can't make default plex patterns. I'm guessing that need dynamically change the patterns based on plex URLs from config.

glensc:
for troubleshooting `config` subcommand also prints merged policy rules

PR #1248, opened 2022-12-01
Fix has_plex_token check

glensc:
Fix https://github.com/Taxel/PlexTraktSync/discussions/1241#discussioncomment-4283101


glensc:
The initial setup passes now. So the problem reported is fixed. But will fail with connection refused on localhost error because it will attempt to connect to empty "default" server. But second run will run with the user configured server.

PR #1264, opened 2022-12-07
Refactor: Use Timer class in TraktBatch

glensc:


glensc:
Issue encountered: If the batch delay is not set (`--batch-delay=0`) there's no code path that will submit the data. The problem is already before this pull request.

PR #1277, opened 2022-12-11
Feature: Add --expire flag to cache command

glensc:


glensc:
@JWCook what I'm doing wrong here that the cache delete doesn't work?

this is cache dump:

```
➜ ./cache.sh https://api-v2launch.trakt.tv/users/me
// request: GET https://api-v2launch.trakt.tv/users/me, response: 200 (103 bytes), created: 2022-12-11 20:32:28 EET, expires: N/A (fresh)
{
  "username": "glen666",
  "private": true,
  "name": "glen",
  "vip": false,
  "vip_ep": false,
  "ids": {
    "slug": "glen666"
  }
}
```

this call should delete the cache entry, but repeated call shows that it still can find the cached entry

```
➜ ./cache.sh https://api-v2launch.trakt.tv/users/me --expire
Expiring: ['f034d5cd4f9c792c']
➜ ./cache.sh https://api-v2launch.trakt.tv/users/me --expire
Expiring: ['f034d5cd4f9c792c']
➜
```

glensc:
with `session.cache.delete(*delete_keys)` (e32fb4023cefda42027053c19d402ca86d867230) still same:

```
➜ ./cache.sh https://api-v2launch.trakt.tv/users/me --expire
Deleting: ['f034d5cd4f9c792c']
Deleting 1 responses
➜ ./cache.sh https://api-v2launch.trakt.tv/users/me --expire
Deleting: ['f034d5cd4f9c792c']
Deleting 1 responses
```

JWCook:
It looks like it's an incorrectly worded log message. It logs that after collecting cache keys (if other settings like `expired` were passed) and before deleting them, without verifying if all of those keys actually exist. It should say something more like `Deleting up to {x} cache keys` instead. I'll change that.

Here's an example to show that delete is actually working as expected:
```python
import logging
from requests_cache import CachedSession

logging.basicConfig(level='DEBUG')

session = CachedSession(use_temp=True)
session.cache.clear()
r1 = session.get('https://httpbin.org/get')
r2 = session.get('https://httpbin.org/json')

session.cache.delete(r1.cache_key, r2.cache_key)
print(f'Cache size: {len(session.cache.responses)}')
session.cache.delete(r1.cache_key, r2.cache_key)
```

The second call will log `Deleting 2 responses`, but the items have already been deleted.

In your example, after running `cache.sh`, can you check whether that response still exists in the cache? You can check by key or by URL:
```python
print(session.cache.contains('f034d5cd4f9c792c'))
print(session.cache.contains(url='https://api-v2launch.trakt.tv/users/me'))
```

glensc:
with 623433d7d6af0e316aa13143a540b4ba1b09132e:

```
+ exec ./plextraktsync.sh cache https://api-v2launch.trakt.tv/users/me --expire
Deleting: ['fef0015d6ba899d5']
True
True
```

I've omitted actual delete in that run.

JWCook:
Hmm, that's unexpected. Do you get the same results if you pass that URL directly to `delete()`?

```python
session.cache.delete(urls=[url])
```

And what happens if you use `del`?
```
del session.cache.responses[cache_key]
```

glensc:
1. `session.cache.delete(urls=[url])` works, as it's the code I copied into https://github.com/Taxel/PlexTraktSync/commit/623433d7d6af0e316aa13143a540b4ba1b09132e
2. `del` results `raise KeyError`, because the computed `cache_key` is wrong.

glensc:
Added some debug, seems problem is due difference to `request.body` being `b'None'` or `''`:

```py
def create_key
...
    key_parts = [
        request.method or '',
        request.url,
        request.body or '',
        request_kwargs.get('verify', True),
        *get_matched_headers(request.headers, match_headers),
    ]
    print(f"{len(key_parts)}: {key_parts}")
```

```py
        print(session.cache.create_key(Request('GET', response.url)))
        print(session.cache.create_key(response.request))
```

```
4: ['GET', 'https://api-v2launch.trakt.tv/', '', True]
d2ad978320e652d6
4: ['GET', 'https://api-v2launch.trakt.tv/', b'None', True]
1d9bf8095870ca55
```

what's in `request.body` property anyway? as the actual body is in `request.content` property.

glensc:
for the sake of this PR, I will just use `session.cache.delete(urls=[url])`, the b'None' problem can be solved elsewhere.

PR #1279, opened 2022-12-11
Refactor: Add plex cache policy patterns

glensc:
Continue https://github.com/Taxel/PlexTraktSync/pull/1221, but for plex urls.

```sh
git grep --function-context -C2 @nocache
```

Docs:
- https://requests-cache.readthedocs.io/en/stable/user_guide/expiration.html#url-patterns

### DIFF
--- a/plextraktsync/config/HttpCacheConfig.py
+++ b/plextraktsync/config/HttpCacheConfig.py
@@ -34,6 +34,53 @@ class HttpCacheConfig:
         "*.trakt.tv/users/*/watchlist/shows": DO_NOT_CACHE,
         "*.trakt.tv/users/likes/lists": DO_NOT_CACHE,
         "*.trakt.tv/users/me": DO_NOT_CACHE,
+
+        # Online Plex patterns
+        "metadata.provider.plex.tv/library/metadata/*/userState": DO_NOT_CACHE,
+        "metadata.provider.plex.tv/library/metadata/*?*includeUserState=1": DO_NOT_CACHE,
+        "metadata.provider.plex.tv/library/metadata/*": LONG_EXPIRY,
+        "metadata.provider.plex.tv/library/sections/watchlist/all": DO_NOT_CACHE,
+        # plex account
+        "plex.tv/users/account": DO_NOT_CACHE,
+
+        # Plex patterns
+        # Ratings search
+        "*/library/sections/*/all?*userRating%3E%3E=-1*": DO_NOT_CACHE,
+        # len(PlexLibrarySection)
+        "*/library/sections/*/all?includeCollections=0&X-Plex-Container-Size=0&X-Plex-Container-Start=0": DO_NOT_CACHE,
+        # __iter__(PlexLibrarySection)
+        "*/library/sections/*/all?includeGuids=1": DO_NOT_CACHE,
+        # find_by_title
+        "*/library/sections/*/all?includeGuids=1&title=*": DO_NOT_CACHE,
+        # fetch_item, fetch_items
+        "*/library/sections/*/all?*": DO_NOT_CACHE,
+        "*/library/sections/*/collections": DO_NOT_CACHE,
+        # library_sections
+        "*/library/sections": DO_NOT_CACHE,
+
+        # reloads
+        "*/library/metadata/*?*include*": DO_NOT_CACHE,
+        # episodes
+        "*/library/metadata/*/allLeaves": DO_NOT_CACHE,
+        # find_by_id
+        "*/library/metadata/*": DO_NOT_CACHE,
+        # mark played, mark unplayed
+        "*/:/scrobble?key=*&identifier=com.plexapp.plugins.library": DO_NOT_CACHE,
+        "*/:/unscrobble?key=&&identifier=com.plexapp.plugins.library": DO_NOT_CACHE,
+        # playlists
+        "*/playlists?title=": DO_NOT_CACHE,
+        "*/playlists/*/items": DO_NOT_CACHE,
+        "*/library": DO_NOT_CACHE,
+        # history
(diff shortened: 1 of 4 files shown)


glensc:
An easy way to find the `@nocache` patterns is to run full sync and grep logs for:

```
grep -o 'Skipping cache write for URL: https://.*plex.direct.*' plextraktsync.log
```

glensc:
there's a problem, can't set rule for root URL (used for "version" and "updated_at" calls), as it will match everything 

```py
        # version, updated_at
        "*/": DO_NOT_CACHE,
```

glensc:
could use  `filter_fn` to filter root url:
- https://requests-cache.readthedocs.io/en/stable/user_guide/filtering.html#custom-cache-filtering

twolaw:
Aren't login urls needed too ?

```
*.trakt.tv/oauth
plex.tv/api/resources
plex.tv/api/users
...
```

glensc:
> Aren't login urls needed too ?
> 
> ```
> *.trakt.tv/oauth
> plex.tv/api/resources
> plex.tv/api/users
> ...
> ```

1. I haven't finished urls about plex, but sure you can add your commits here
2. do you have actual issue with trakt URLs, please submit PR to fix.

with trakt pull request, I went every method and looked what requests they make and converted `@nocache` to rule. so URLs that did not have `@nocache` before did not get rule added, so they should continue to behave like before.

there's slight possibility if some method under `@nocache` made multiple requests based on conditions, then those I might have missed.

plextraktsync/config/HttpCacheConfig.py
```diff
@@ -30,6 +30,28 @@ class HttpCacheConfig:
...
         "*.trakt.tv/users/likes/lists": DO_NOT_CACHE,
         "*.trakt.tv/users/me": DO_NOT_CACHE,
+
+        # Plex patterns
+        # Ratings search
+        "*/library/sections/*/all?*userRating%3E%3E=-1*": DO_NOT_CACHE,
+        # fetch_item
+        "*/library/sections/*/all?*": DO_NOT_CACHE,
+        "*/library/sections/*/collections": DO_NOT_CACHE,
+        "*/library/metadata/*": DO_NOT_CACHE,
```
twolaw, line 40: I think `metadata.provider.plex.tv/library/metadata/*` should be cached because there is no user data in response.
But `metadata.provider.plex.tv/library/metadata/*/userState` must remain not cached.

glensc: Added dc07f64f, but perhaps should have still some cache expiry rather `NEVER_EXPIRE`? how much?

twolaw: If we could have a random value, that'd be perfect.
It would allow to spread out the requests.
E.g. 3 month +random(0-30days)

glensc: b667c0eb, 5f2243b1.

but how to test that? enable all watchlist related options and like some online movie?

glensc: there's no `metadata.provider.plex.tv/library/metadata/*/userState`, but rather `https://metadata.provider.plex.tv/library/metadata/*?*includeUserState=1` and it includes view counts etc.

but I don't think we use those values, only test for guids

glensc:
Basically, all methods were manually verified and moved `@nocache` to URL patterns.

but could be that I didn't test all agent URLs, I'm still using Legacy Agent for shows due performance reasons.

`PlexServerConnection.connect` remains with `@nocache` because it's pattern would be `"*"`, making it apply for all urls.

plextraktsync/config/HttpCacheConfig.py
```diff
@@ -30,6 +34,52 @@ class HttpCacheConfig:
         "*.trakt.tv/users/*/watchlist/shows": DO_NOT_CACHE,
         "*.trakt.tv/users/likes/lists": DO_NOT_CACHE,
         "*.trakt.tv/users/me": DO_NOT_CACHE,
+
+        # Online Plex patterns
+        "metadata.provider.plex.tv/library/metadata/*/userState": DO_NOT_CACHE,  # XXX bogus?
+        "metadata.provider.plex.tv/library/metadata/*?*includeUserState=1": DO_NOT_CACHE,
+        "metadata.provider.plex.tv/library/metadata/*": LONG_EXPIRY,
```
twolaw, line 41: I suggest to simply use `NEVER_EXPIRES` or `timedelta(weeks=x)`

glensc: will go with timedelta without rand

twolaw:
Heavily tested I assume 👼

glensc:
> Heavily tested I assume 👼

It's not git tagged yet. but fingers crossed!

it already solved one issue for me that `inspect <season_id>` cached the result and episode.media.part.exists leading to wrong cached value.

twolaw:
> fingers crossed

Come on....
Don't tag before tests.

glensc:
what do you mean? I've been testing it and running it for a while. and I haven't made tag, that's what I said, it's not tagged yet.

twolaw:
I thought you didn't test and only cross finger as a quality process.

glensc:
I wrote how I tested:
- https://github.com/Taxel/PlexTraktSync/pull/1279#issuecomment-1358839429

you didn't reply anything to it. so I assumed you were fine with this. and plexapi fixes got released too.

plextraktsync/config/HttpCacheConfig.py
glensc: randomize expire request:
- https://github.com/requests-cache/requests-cache/issues/768

PR #1287, opened 2022-12-13
Feature: Add --no-cache flag to inspect command

glensc:
Simplified version of #1141


plextraktsync/factory.py
```diff
@@ -80,9 +80,17 @@ def server_config(self):
     def session(self):
         from requests_cache import CachedSession
 
+        if self.run_config.cache:
+            urls_expire_after = self.config.http_cache.urls_expire_after
+        else:
+            from requests_cache import DO_NOT_CACHE
+            urls_expire_after = {
+                "*": DO_NOT_CACHE,
```
glensc, line 88: maybe there's a better way for t his. but since this is for debugging, it doesn't matter much.

PR #1312, opened 2022-12-22
HttpCache: Filter plex urls

glensc:
Trickery to be able to add rule for plex server root url.

### DIFF
--- a/plextraktsync/config/HttpCacheConfig.py
+++ b/plextraktsync/config/HttpCacheConfig.py
@@ -104,7 +104,7 @@ class HttpCacheConfig:
         # system_account
         "*/accounts": DO_NOT_CACHE,
         # version, updated_at
-        # "*/": DO_NOT_CACHE,
+        "*.plex.direct:*/": DO_NOT_CACHE,
     }
 
     @property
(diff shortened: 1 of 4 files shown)


glensc:
With this hack, last `@nocache` is dropped.

glensc:
this is way too complicated. maybe better approach would be to change default policy

glensc:
or why not just add only this chunk:

```py
        "*.plex.direct:*/": DO_NOT_CACHE,
```

glensc:
> or why not just add only this chunk:
> 
> ```python
>         "*.plex.direct:*/": DO_NOT_CACHE,
> ```

Because not all plex URLs are in that domain.

glensc:
Maybe using regexp pattern is simpler:
- https://github.com/requests-cache/requests-cache/blob/2c143a395d27eebd599f628df03f190175e67818/requests_cache/policy/expiration.py#L116

glensc:
No longer needed as of:
- https://github.com/Taxel/PlexTraktSync/pull/1870

More specifically:
- https://github.com/Taxel/PlexTraktSync/commit/467333c8d5b9efef0c2f597d314175d645876d5c

PR #1317, opened 2022-12-22
Add year to watchlist add/remove message

glensc:
This is not much useful:

```
INFO     Removing 'Innocence' from Plex watchlist
```

Currently in Plex global search there are 6 movies with that exact title, so which ones were removed?


glensc:
Personally, I would just put `{m}` there, then it has all the details needed by the media object.

I was trying to find the item because the script removed it, but could not find the item in plex at all. so still don't know what could it have been

twolaw (approved):

PR #1332, opened 2023-01-02
Add TYPE_CHECKING to modules

glensc:
Distinguish code import and types imports.


glensc:
```
$  git grep -l 'from typing import TYPE_CHECKING'|xargs grep -c 'from __future__ import annotations'|grep -v :1 -c
0
```

PR #1346, opened 2023-01-09
Feature: Add --sync=watchlist option

glensc:
Allows syncing only watchlist:

```
$ plextraktsync sync --sync=watchlist
```

does not work when `watchlist_as_playlist` is enabled.

syncing plain watchlist, does not need to walk through libraries, so can work fine in partial mode or no library walk mode at all.

### DIFF
--- a/plextraktsync/sync.py
+++ b/plextraktsync/sync.py
@@ -153,7 +153,7 @@ class Sync:
             for show in walker.walk_shows(shows, title="Syncing show ratings"):
                 self.sync_ratings(show, dry_run=dry_run)
 
-        if self.sync_wl or self.config.sync_liked_lists:
+        if walker.config.walk_watchlist and (self.sync_wl or self.config.sync_liked_lists):
             if is_partial:
                 logger.warning("Partial walk, watchlist and/or liked list updating was skipped")
             else:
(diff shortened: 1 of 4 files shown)


plextraktsync/sync.py
```diff
@@ -153,15 +153,16 @@ def sync(self, walker: Walker, dry_run=False):
             for show in walker.walk_shows(shows, title="Syncing show ratings"):
                 self.sync_ratings(show, dry_run=dry_run)
 
-        if self.sync_wl or self.config.sync_liked_lists:
+        if self.config.update_plex_wl_as_pl or self.config.sync_liked_lists:
             if is_partial:
-                logger.warning("Partial walk, watchlist and/or liked list updating was skipped")
+                logger.warning("Partial walk, liked list updating was skipped")
```
twolaw, line 158: To be consistent, reformulate to 
`logger.warning("Running partial library sync. Liked lists won't update because it needs full library sync.")`

glensc: if you use github code suggestions would be a lot better to accept changes!

```suggestion
                logger.warning("Running partial library sync. Liked lists won't update because it needs full library sync.")
```

PR #1362, opened 2023-01-14
Cleanup: Unify use of print

glensc:
Remove uses of `click.echo`, Replace uses with Rich `print`. We don't need Python 2 wrappers, and print from Rich does extra coloring, which is nice.

### DIFF
--- a/plextraktsync/commands/cache.py
+++ b/plextraktsync/commands/cache.py
@@ -3,8 +3,6 @@ from __future__ import annotations
 from functools import partial
 from typing import TYPE_CHECKING
 
-import click
-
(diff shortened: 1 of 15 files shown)


glensc:
Some side effect, rich `print` wraps text terminal width, for example `cache` command output is no longer greppable because items are not printed on single line anymore:

```
$ plextraktsync cache --limit=-1|grep N/A
...
expires: N/A (fresh)
```

glensc:
but wrap can be turned off:
- https://github.com/Taxel/PlexTraktSync/pull/1364

PR #1375, opened 2023-01-16
Use multline server name selection in plex login

glensc:
Some experiments with InquirerPy and rich. Looks nice.


glensc:
Currently, this all is a mess, mixing click, inquirerpy, rich. Probably should create some common UI class with methods, like error box, success box, password prompt, etc.

also, rich methods should pass the console instance so it wouldn't use the global

PR #660, opened 2021-12-19
Feature: Add __rich__ method to PlexLibraryItem

glensc:
Requires: https://github.com/Taxel/PlexTraktSync/pull/659

PR #1447, opened 2023-03-27
Use rich links when printing titles

glensc:
This fixes issues like:
1. titles with aposthrophes continue be colored
2. having title on screen it's difficult to find actual item

now it's clickable links, if terminal supports it.

### DIFF
--- a/plextraktsync/media.py
+++ b/plextraktsync/media.py
@@ -9,6 +9,7 @@ from trakt.errors import TraktException
 from plextraktsync.decorators.cached_property import cached_property
 from plextraktsync.factory import logger
 from plextraktsync.trakt.TraktLookup import TraktLookup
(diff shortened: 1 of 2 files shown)


glensc:
Could add some custom log level:
- https://stackoverflow.com/questions/2183233/how-to-add-a-custom-loglevel-to-pythons-logging-facility/35804945#35804945

Not sure if  that's worth it

twolaw:
The same could be applied for sections titles. Currently there is no formatting at all.

`INFO     Films processed in 14.0 seconds`

twolaw (approved):

glensc:
maybe this does it automatically:
- https://github.com/Taxel/PlexTraktSync/pull/660

PR #1448, opened 2023-03-29
Feature: Use rich links in print_plan

glensc:
See https://github.com/Taxel/PlexTraktSync/pull/1447#issuecomment-1487223403


glensc:
For other items in WalkPlan:
- https://github.com/Taxel/PlexTraktSync/blob/f33ef68568c5ed5814065e4e381cdaad5c1293b5/plextraktsync/walker.py#L90-L92

need to refactor to use PlexLibraryItem there first.

PR #1475, opened 2023-05-02
Refactor: Playlists refactoring

glensc:
The intention is to print "Added to list" message only if it's really missing from the list

### DIFF
--- a/plextraktsync/media.py
+++ b/plextraktsync/media.py
@@ -81,6 +81,10 @@ class Media:
     def trakt_id(self):
         return self.trakt.trakt
 
+    @cached_property
+    def plex_key(self):
+        return self.plex.key
+
     @property
     def trakt_url(self):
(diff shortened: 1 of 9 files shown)


simonc56:
> The intention is to print "Added to list" message only if it's really missing from the list

Tested, it doesn't seem to work.
Script always print 'Added to list' for all items of a list, even if items were already in the Plex list.

simonc56:
Suggestions for lists sync logs : 
- make Plex lists names clickable in terminal output to open them in Plex
- show beautiful Title with year instead of <Movie:76:The-Lord-of-the -ri> when added to list

simonc56:
Tested on wrong branch... will test again later.

simonc56:
Tested successfully ✅

simonc56 (approved):

Issue #1489, opened 2023-05-24
Feature request: Prefer most recent ratings OR skip setting rating for items with existing ratings


RileyXX:
## Issue:

PlexTraktSync is overwriting Trakt ratings with outdated ratings from Plex. 

## Feature expectation: 

The expectation of this feature would be, when syncing items the script should prefer the rating that was submitted most recently and sync both ways accordingly. If necessary, the outdated rating should be overwritten. 

If it is not possible to check when the rating was submitted, then the script should check if there is already a rating for that item, otherwise skip setting the rating for that item. 

## Adding the feature as an optional config setting:

This could be implemented as an optional config setting incase the original method of overwriting ratings where Plex is the source of ratings is preferred by the user. 

## Why this feature should be added:

Plex uses a 5 star rating system and Trakt uses a 10 star system. So for example, if you left a 3 star rating in Plex then synced with PlexTraktSync. Afterwards you go into Trakt and saw it is now 7/10. Then you decide to change the rating to 6/10 on Trakt. Next time you run PlexTraktSync it will be overwritten again to 7/10 on Trakt. This is just an example, and actual rating value might vary. 

I have other programs that are syncing my ratings from Trakt to IMDB etc so I like to keep my ratings in Trakt as updated and accurate as possible. But PlexTraktSync is overwriting my most recent Trakt ratings with outdated ratings from Plex. Would like to see this feature implemented so I can sync ratings using PlexTraktSync without new ratings being overwritten.

RileyXX:
Would the following modification work in [sync.py](https://github.com/Taxel/PlexTraktSync/blob/299113e528534e844655fb98f26306b1a5927a7e/plextraktsync/sync.py#LL130C1-L145C30) to skip setting rating for items with existing ratings?

````python
    def sync_ratings(self, m: Media, dry_run=False):
        if not self.config.sync_ratings:
            return

        if m.plex_rating == m.trakt_rating:
            return

        if m.plex_rating is not None and m.trakt_rating is None and self.config.plex_to_trakt["ratings"]:
            logger.info(f"Rating {m.title_link} with {m.plex_rating} on Trakt", extra={"markup": True})
            if not dry_run:
                m.trakt_rate()
        elif m.trakt_rating is not None and m.plex_rating is None and self.config.trakt_to_plex["ratings"]:
            logger.info(f"Rating {m.title_link} with {m.trakt_rating} on Plex", extra={"markup": True})
            if not dry_run:
[...]

PR #1493, opened 2023-06-03
Modify rating sync logic. Only rate items with missing rating. Don't overwrite existing ratings.

RileyXX:
### Description:
Fix for issue: https://github.com/Taxel/PlexTraktSync/issues/1489. Small modification for rating sync logic in sync.py. Only rate items with missing rating. Don't overwrite existing ratings. I did some tests and this modification does seem to be working as intended.

### New Logic: 
If the Plex rating is set and the Trakt rating is not, the Plex rating is applied to Trakt. Similarly, if the Trakt rating is set and the Plex rating is not, the Trakt rating is applied to Plex.

### Old Logic:
Previously, if there was an existing rating for an item in Plex AND Trakt, the Plex rating would take precedence and the Trakt rating would be overwritten.

### Changes:
In the modified code, the additional conditions for `m.plex_rating is not None and m.trakt_rating is None` and `m.trakt_rating is not None and m.plex_rating is None` are added. Additionally, the comment line was modified to explain the new conditions.

### DIFF
--- a/plextraktsync/sync.py
+++ b/plextraktsync/sync.py
@@ -134,12 +134,12 @@ class Sync:
         if m.plex_rating is m.trakt_rating:
             return
 
-        # If two-way rating sync, Plex rating takes precedence over Trakt rating
-        if m.plex_rating is not None and self.config.plex_to_trakt["ratings"]:
+        # If two-way rating sync, only rate items with missing rating
+        if m.plex_rating is not None and m.trakt_rating is None and self.config.plex_to_trakt["ratings"]:
             logger.info(f"Rating {m.title_link} with {m.plex_rating} on Trakt", extra={"markup": True})
             if not dry_run:
                 m.trakt_rate()
-        elif m.trakt_rating is not None and self.config.trakt_to_plex["ratings"]:
+        elif m.trakt_rating is not None and m.plex_rating is None and self.config.trakt_to_plex["ratings"]:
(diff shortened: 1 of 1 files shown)


RileyXX:
Also I was curious about the following [line](https://github.com/Taxel/PlexTraktSync/pull/1493/commits/41e52e5bde77b1c82fcd8cc824715284bfb33943#diff-cf5f976d9877950e58f01982812825547b3aa992b5536ba773b2386fa1b84d1fR134) in sync.py: 
```python
        if m.plex_rating is m.trakt_rating:
            return
```
Is there a reason for using the `is` operator rather than the `==` operator? I haven't looked to deeply into the code so I left this unchanged in my pull request. But I was just curious if this should be changed or left as is.
```python
        if m.plex_rating == m.trakt_rating:
            return
```

RileyXX:
Closed in favor of better implementation in new pull request

glensc:
Don't create pull requests from your `main` branch. You will run into issues. always create new branch

glensc:
I'm guessing `is` vs `==` is because the value can be `None` and `None` should be compared with `is`.

However, both seem to work.

```
$ python
Python 3.11.3 (main, Apr 10 2023, 18:26:05) [Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> None is None
True
>>> None == None
True
>>> 
```

you could find the original commit that added it. but it's a long journey, that part has been rewritten multiple times.

PR #1496, opened 2023-06-05
Feature: Add support for optional rating priority config setting (none, trakt, or plex)

RileyXX:
## Description:
Fix for issue: https://github.com/Taxel/PlexTraktSync/issues/1489. Pull request adds support for optional rating priority config setting. Includes modifications for rating sync logic in `sync.py` and a new setting for `rating_priority` in `config.default.yml`.

## New Config Setting Options:
 Valid values for rating_priority are trakt, plex or none. (default: none)
- none - No rating priority. Existing ratings are not overwritten.
- trakt - Trakt ratings have priority. Existing Plex ratings are overwritten.
- plex - Plex ratings have priority. Existing Trakt ratings are overwritten.

## Old Logic:
Previously, if there was an existing rating for an item in Plex AND Trakt, the Plex rating would take precedence and the Trakt rating would be overwritten. 

## Testing
I have thoroughly tested these changes and have found no conflicts or errors. Everything seems to be working as intended.

### DIFF
--- a/plextraktsync/config.default.yml
+++ b/plextraktsync/config.default.yml
@@ -64,11 +64,12 @@ sync:
     # If you prefer to fetch trakt watchlist as a playlist instead of
     # plex watchlist, toggle this to true (is read only if watchlist=true)
     watchlist_as_playlist: false
-  # Setting for whether or not ratings from one platform should have priority. Valid values are trakt, plex or none. (default: none)
+  # Setting for whether ratings from one platform should have priority.
+  # Valid values are trakt, plex or none. (default: plex)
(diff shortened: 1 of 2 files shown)


glensc:
1. default value must be same as before (which is plex takes preference)
2. code should be simplified, with lots of copy-paste making it challenging to maintain the code in the future.

glensc:
Extra feature: See when ratings differ (Display info). Currently if they differ they are silently ignored.

RileyXX:
Thanks for taking the time to modify and implement this! I wanted to simplify it but the logic was making my brain melt lol. The way you did it is clever. Nicely done!

PR #1514, opened 2023-06-27
reorganize ClickException in Trakt AuthException

twolaw:
Exception in Exception in Exception is too noisy

No Trakt connection -> stop the script with CRITICAL error

### DIFF
--- a/plextraktsync/trakt/TraktApi.py
+++ b/plextraktsync/trakt/TraktApi.py
@@ -52,8 +52,7 @@ class TraktApi:
         try:
             return trakt.users.User("me")
         except (OAuthException, ForbiddenException) as e:
-            logger.fatal(f"Trakt authentication error: {str(e)}")
-            raise ClickException(e)
+            raise ClickException(f"Trakt authentication error: {str(e)}")
 
     @cached_property
     @rate_limit()
--- a/plextraktsync/cli.py
+++ b/plextraktsync/cli.py
@@ -2,6 +2,7 @@ from functools import wraps
 from os import environ
 
 import click
+from click import ClickException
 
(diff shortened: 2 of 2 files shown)


glensc:
what is the problem you trying to solve? show some output before/after.

plextraktsync/trakt/TraktApi.py
```diff
@@ -53,7 +53,7 @@ def me(self):
             return trakt.users.User("me")
         except (OAuthException, ForbiddenException) as e:
             logger.fatal(f"Trakt authentication error: {str(e)}")
-            raise ClickException(e)
+            sys.exit(1)
```
glensc, line 56: sys.exit in a class? that's a no!

class methods should throw, and this is what it is already doing now.

twolaw: Ok but at the moment we have 3 cascaded exceptions throwed if bad trakt authentication :
OAuthException > ClickException > TypeError

Isn't that bad practice ?

My first fix to remove TypeError was :
```diff
- raise ClickException(e)
+ raise ClickException(str(e))
```

glensc: you should add also to cli.py:

```py
            except ClickException as e:
                raise ClickException(f"Error running {name} command: {e}")
```

before:
```py
            except Exception as e:
                from plextraktsync.factory import logger
                logger.exception(e)

```

so, that `ClickException` is exiting without printing the trace (too noisy).

twolaw:
Fixed.
No more exceptions cascade when Trakt authentication fail.
Just a CRTITICAL error.

glensc (approved):

PR #1532, opened 2023-07-21
Fetch episodes using library search to include guids without reloading

JonnyWong16:
## Description

Fetch episodes using library search to include guids without reloading every episode object. This should *significantly* increase the speed for syncing tv shows.

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated the docstring for new or existing methods
- [ ] I have added tests when applicable

### DIFF
--- a/plextraktsync/config/HttpCacheConfig.py
+++ b/plextraktsync/config/HttpCacheConfig.py
@@ -86,7 +86,7 @@ class HttpCacheConfig:
         # reloads
         "*/library/metadata/*?*include*": DO_NOT_CACHE,
         # episodes
-        "*/library/metadata/*/allLeaves": DO_NOT_CACHE,
+        "*/library/sections/*/all?includeGuids=1&type=4*": DO_NOT_CACHE,
         # find_by_id
         "*/library/metadata/*": DO_NOT_CACHE,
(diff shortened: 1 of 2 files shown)


JonnyWong16:
Side question: Can you just skip the show object and process episodes directly? e.g. Using `plex.library.section("TV Shows").searchEpisodes()` instead of `plex.library.section("TV Shows").all()`. This would cut it down to a single API call to retrieve all episodes in the library instead of one API call per show.

twolaw:
> This should significantly increase the speed for syncing tv shows.

Indeed, it improves speed a lot. Thanks!

> Side question: Can you just skip the show object and process episodes directly? e.g. Using plex.library.section("TV Shows").searchEpisodes() instead of plex.library.section("TV Shows").all(). This would cut it down to a single API call to retrieve all episodes in the library instead of one API call per show.

Yes I think we could do that.

twolaw (approved):

Issue #976, opened 2022-07-13
Feature request: Change processing rate calculation for shows sections


simonc56:
### Feature description

Show sections : IMHO items/s should be calculated with episodes/s not with shows/s.

### Use case

When processing **movies** sections, the rate of processed items per second is simple: 1 item = 1 movie.
But for **shows** sections, the rate of processed items per second is (according to me) not correct because it is calculated based on 1 item = 1 show. Shows don't have same quantity of episodes, some have ~10 episodes and some have hundreds so **the actual calculation is not a good measure of script speed**. It should be 1 item = 1 episode.

![image](https://user-images.githubusercontent.com/10365746/178788212-239d1418-9d14-4837-9869-f4ab233793ab.png)

On my personal library, I have about 163 items/s for movies and 1 item/s for shows.

I know that movies sections and shows sections are not processed the same way, but I expect the items/s rate to be more or less consistent between sections to be a usefull value.

### Workarounds

Not that I know

### Plan to implement

I don't know how to implement this.
I'm not skilled at `yield` python things and will need help if i want to try to implement myself.

glensc:
Changing the increment to emit episodes, is doable, but then the percentage will be unknown.

glensc:
other option is to use multiple or sub-progressbars:
- https://github.com/Textualize/rich#rich-library

Jopo-JP:
Maybe two progress bars?
One for all the show and one for that specifiy show and their epiosdes?

Example:
Processing Shows 80%   ━━━━━━━━━━━━━━━━━━━━━━━━━   80/100[ 0:03:00 < 0:00:00 , 1 it/s ]
Processing Big Bang Theory 100%  ━━━━━━━━━━━━━━━━━━━━   20/100 [ 0:00:05 < 0:00:00 , 20 it/s ]

glensc:
@Jopo-JP that's precisely what I said:
- https://github.com/Taxel/PlexTraktSync/issues/976#issuecomment-1184811444

Jopo-JP:
> @Jopo-JP that's precisely what I said:
> - https://github.com/Taxel/PlexTraktSync/issues/976#issuecomment-1184811444

Well this is at least the best way to handle it.

Otherwise pre scanning all episodes or counting it before scanning could take a while to start the initial scan.

I don't know if there is a API which will print all episode not just the shows.

[...]

PR #1578, opened 2023-09-11
Refactor: Implement shows sync using episodes

glensc:
Implement shows sync using episodes search:
- https://github.com/Taxel/PlexTraktSync/issues/976#issuecomment-1648004190

<del>Requires support from PlexApi side: - https://github.com/pkkid/python-plexapi/pull/1243</del>

### DIFF
--- a/plextraktsync/media.py
+++ b/plextraktsync/media.py
@@ -89,8 +89,7 @@ class Media:
     @property
     def show(self) -> Media | None:
         if self._show is None and self.mf and not self.plex.is_discover:
-            # TODO: fetch show for discover items
-            ps = self.plex_api.fetch_item(self.plex.item.grandparentRatingKey)
+            ps = self.plex.show
             ms = self.mf.resolve_any(ps)
             self._show = ms
 
--- a/plextraktsync/media.py
+++ b/plextraktsync/media.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 from plexapi.exceptions import PlexApiException
 from requests import RequestException
 from trakt.errors import TraktException
+from trakt.tv import TVShow
 
 from plextraktsync.factory import logger
 from plextraktsync.trakt.TraktLookup import TraktLookup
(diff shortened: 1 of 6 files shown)


glensc:
Actually for me the syncing is even slower:
1. Using legacy TV agent
2. 30 sec (old code) vs 45 sec (new code)

reason being is that extra api calls are still there:
```
Made query /library/metadata/7632
```

glensc:
timing again:
- this branch: Completed full sync in 2 min 9.3 seconds
- master branch: Completed full sync in 33.6 seconds

So, the new code is actually slower.

JonnyWong16:
I think the slowdown is because it is now querying Trakt for every episode since it doesn't create the `TraktLookup` table anymore.

It's faster on the Plex side, but slower on the Trakt side.

glensc:
I haven't verified this, but I think it's also slower now because needs to fetch every plex show separately. 

can probably can be solved by batching show fetches as well.

glensc:
Added plex shows preloading and caching show object for next episode of same show. Timing looks promising.

scanning a library with 227 shows, 5180 episodes with hot cache. the library is mostly healthy (few missing matches) and using legacy agent. there's also bunch of reloading episode/show for year and userRating, that slows down as well.

- this branch: 12.6-13.9 seconds            
- main branch: 13.7-17.0 seconds

glensc:
I'll go on with merge then

simonc56:
Tested successfully.
- seems faster than before
- strange to Process each tvshows library twice

glensc:
it's preloading tvshows in first run. do you have recommendations for it? make it silent? change display?

also, it's processing fast the first round right?

can you also post your timings of now and then?

simonc56:
After more tests, i see no time difference between `0.28.4` and `0.28.5` when I sync ratings and watched status.
With those settings, speed is 100\~140 episodes/sec and 50\~80 movies/sec.


image


I suggest to slightly change display like that : 

image

glensc:
what is the difference? if you made code changes so just submit them?

simonc56:
I just draw it on paint.

image

PR #1635, opened 2023-11-26
Re-use ACCOUNT token for plex-token

glensc:
Re-use ACCOUNT token for plex-login to prevent "Signed in from new location" e-mail notifications.

### DIFF
--- a/plextraktsync/commands/plex_login.py
+++ b/plextraktsync/commands/plex_login.py
@@ -196,7 +196,13 @@ def login(username: str, password: str):
         if not Confirm.ask(PROMPT_PLEX_RELOGIN, default=True):
             return
 
-    account = myplex_login(username, password)
+    config = factory.config
+
+    if config["PLEX_ACCOUNT_TOKEN"]:
+        account = MyPlexAccount(username=username, token=config["PLEX_ACCOUNT_TOKEN"])
+    else:
+        account = myplex_login(username, password)
+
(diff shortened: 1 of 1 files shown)


glensc:
> It prevents the user to login with new credentials?

what it? the pr as a whole or my [comment](https://github.com/Taxel/PlexTraktSync/pull/1635#issuecomment-1826794458)?

simonc56:
> wouldn't it be consistent to always write PLEX_ACCOUNT_TOKEN if it's known it's account token?

In my memory, PLEX_ACCOUNT_TOKEN = PLEX_TOKEN when the user is the PMS admin.

See description here :
https://github.com/Taxel/PlexTraktSync/discussions/1288#discussioncomment-4397824

glensc:
The idea behind this PR was to able to use plex-login without getting email notification I signed in from new place.

I have to verify that sign in with account token doesn't send this email. i think it will send email regardless of sign method and if the signed in ip is new.

if so, this pr serves no actual purpose.

glensc:
Not needed

glensc:
>  This defeats the purpose of plex-login command, I think.

Added with --use-token flag only:
- [`46e3485` (#2126)](https://github.com/Taxel/PlexTraktSync/pull/2126/commits/46e34855c71c795fde4a0b65f2e8d4adbf99ac3f)

PR #1654, opened 2023-12-05
Feature: Add systemd logger

glensc:
Try systemd logging:
- https://trstringer.com/systemd-logging-in-python/


glensc:
Not sure how to setup the extras, so `systemd` would not be attempted to install by default:
- https://github.com/pypa/pipenv/issues/6034

glensc:
and systemd not available on alpine:
```
/ # apk add systemd
ERROR: unable to select packages:
  systemd (no such package):
    required by: world[systemd]
```

so, need to setup the marker to install only on non-alpine linux, or libc linuxes or somehow make it optional.

glensc:
here's pep what could be filtered with markers:
- https://peps.python.org/pep-0496/

glensc:
Dump of modifying the journal in display mode

image

glensc:
@simonc56 what's your interest in this? i.e. why the eyes?

glensc:
A way to set is `optional = True`:
```
tree_sitter_languages = {version = ">=1.7.0", optional = true}
```

- https://github.com/Textualize/textual/pull/3398/files

Issue #1631, opened 2023-11-14
plex 500 internal error with index=None


digitalgp:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists
- [X] I have tried downgrading to find version that can be used as a workaround

### The problem

i got this error in the plex console when running plextraktsync. 

```
Got exception from request handler: sqlite3_statement_backend::prepare: no such column: None for SQL: select count(*) from (select distinct(metadata_items.id) from metadata_items join metadata_items as parents on metadata_items.parent_id=parents.id join metadata_items as children on children.parent_id=metadata_items.id where metadata_items.library_section_id in (1) and (metadata_items.`index`=None and metadata_items.metadata_type=3 and metadata_items.parent_id=66153) )
```

  anyone seen this? and have a fix? is it the plextraktsync thats broken?

### Error trace / logs

```text
2023-11-14 22:59:45,270 ERROR[PlexTraktSync]:(500) internal_server_error; https://192-168-10-20.29069781034849ab98845e008b75f47c.plex.direct:32400/library/metadata/66153/children?excludeAllLeaves=1&index=None <html><head><title>Internal Server Error</title></head><body><h1>500 Internal Server Error</h1></body></html>
Traceback (most recent call last):
  File "/app/plextraktsync/cli.py", line 26, in wrap
    cmd(*args, **kwargs)
  File "/app/plextraktsync/commands/sync.py", line 68, in sync
    runner.sync(walker=w, dry_run=config.dry_run)
  File "/app/plextraktsync/sync.py", line 79, in sync
    for episode in walker.find_episodes():
  File "/app/plextraktsync/walker.py", line 303, in find_episodes
    yield from self.episode_from_show(show)
  File "/app/plextraktsync/walker.py", line 343, in episode_from_show
    me = self.mf.resolve_any(pe, show)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/plextraktsync/media.py", line 252, in resolve_any
    m = self.resolve_guid(guid, show)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/plextraktsync/media.py", line 260, in resolve_guid
    logger.warning(f"{guid.pm.item}: Skipping {guid} because provider {guid.provider} has no external Id")
                     ^^^^^^^^^^^^^^
[...]

PR #1657, opened 2023-12-06
Fix: Skip dumping guid.pm.item on unmatched items

glensc:
This mitigates the problem with fetching `index=None` from plex api

See:
- https://github.com/Taxel/PlexTraktSync/issues/1631


simonc56:
I think we can wait for Plexapi fix instead of this.

glensc:
It happened to me, a server whose version I can't control. and it happened for unmatched items. and the library was legacy tv show type.

simonc56:
Ok go ahead then, we'll reverse this PR when the plexapi fix will be released.

PR #1678, opened 2023-12-27
Feature: Retry movie search for mini series

glensc:
Quite typical to put mini-series as movie to the library, as plex recognizes both. but trakt only has them as tv series.

If the match is not found,  try a tv series and print warning if found, so users would know to fix this.


simonc56:
>Quite typical to put mini-series as movie to the library

Typical ? Can you give examples ?

glensc:
> Typical ? Can you give examples ?

-  https://en.wikipedia.org/wiki/It_(miniseries)

simonc56 (approved):

PR #1679, opened 2023-12-27
Display title of unmatched guid in log

simonc56:
sort of reverse of PR #1657

### DIFF
--- a/plextraktsync/media.py
+++ b/plextraktsync/media.py
@@ -255,12 +255,12 @@ class MediaFactory:
 
     def resolve_guid(self, guid: PlexGuid, show: Media = None):
         if guid.provider in ["local", "none", "agents.none"]:
-            logger.warning(f"Skipping {guid} because provider {guid.provider} has no external Id")
+            logger.warning(f"{guid.title_link}: Skipping {guid} because provider {guid.provider} has no external Id", extra={"markup": True})
 
             return None
 
         if guid.provider not in ["imdb", "tmdb", "tvdb"]:
-            logger.error(f"Unable to parse a valid provider from {guid}")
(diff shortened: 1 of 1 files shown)


plextraktsync/media.py
```diff
@@ -255,12 +255,12 @@ def resolve_any(self, pm: PlexLibraryItem, show: Media = None) -> Media | None:
 
     def resolve_guid(self, guid: PlexGuid, show: Media = None):
         if guid.provider in ["local", "none", "agents.none"]:
-            logger.warning(f"Skipping {guid} because provider {guid.provider} has no external Id")
+            logger.warning(f"{guid.title_link}: Skipping {guid} because provider {guid.provider} has no external Id")
```
glensc, line 258: seems you didn't test this?

if you use title link you must also enable markdown in logger:
- https://github.com/Taxel/PlexTraktSync/pull/1673

glensc: do you even need `{guid}` if you use `{guid.title_link}`?

simonc56: Shame on me I didn't test this.

> do you even need `{guid}` if you use `{guid.title_link}`?

Yes I think it's usefull :

image

glensc: not useful for typical user. not useful for me.  for providing bug report, or diagnose you should run inspect anyway.

so, when I want to troubleshoot, I click the plex URL and run inspect of the URL or try to do fix match. i have never needed to copy the 845 since the titles became clickable.

and for consistency should also update message changed here (does not contain `{guid}` anymore):
- https://github.com/Taxel/PlexTraktSync/pull/1673/files

glensc: and indeed, it's useful when sharing output for bug reports:
- https://github.com/Taxel/PlexTraktSync/discussions/1746

PR #1683, opened 2023-12-29
CI: Add release environment for pypi job

glensc:
Taking from pipx:
- https://github.com/pypa/pipx/blob/81c2fa509768715de24a3ba10d46596de980052f/.github/workflows/tests.yml#L91-L93

Should appear something like this:
- https://github.com/pypa/pipx/deployments/release


glensc:
https://github.com/Taxel/PlexTraktSync/deployments/release

PR #1686, opened 2024-01-05
Update trakt baseurl

simonc56:
Updates trakt baseurl because old one is deprecated.

According to [Trakt API](https://twitter.com/traktapi/status/1742976564567106019) :

> The API host name [http://api-v2launch.trakt.tv](https://t.co/9TYHCeYzjb) will be removed on February 1. 
>
>This has been deprecated since 2017. This will only affect very old apps and they need to use [http://api.trakt.tv](https://t.co/0mENbHo7Uq) to continue working beyond February 1.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -242,7 +242,7 @@ curl -sSf \
 ```
 
 ```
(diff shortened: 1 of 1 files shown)


simonc56:
Most important update is in pytrakt module : https://github.com/glensc/python-pytrakt/pull/34

PR #1704, opened 2024-01-07
Fix: Use trakt slug for trakt_url

glensc:
The redirect by id may get to incorrect place:

- https://trakt.tv/movies/96

Redirects wrongly:
- https://trakt.tv/movies/96-2018

Should be:
- https://trakt.tv/movies/princess-mononoke-1997

### DIFF
--- a/plextraktsync/media.py
+++ b/plextraktsync/media.py
@@ -82,7 +82,9 @@ class Media:
 
     @property
     def trakt_url(self):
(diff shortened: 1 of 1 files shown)


simonc56:
I disagree.
slugs have been replaced by id in url on purpose because of issues with changing slugs.

glensc:
as noted in pr body, the numberic id is not reliable. also trakt redirects to new slug if the slug changes. and currently slug is printed in inspect only, so it's short lived, nothing permanent.

Issue #1724, opened 2024-01-10
Rate Limit Exceeded for plextraktsync.trakt.ScrobblerProxy.update since 0.28.2


zeroquinc:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists
- [X] I have tried downgrading to find version that can be used as a workaround

### The problem

I keep getting a Rate Limit Exceeded for plextraktsync.trakt.ScrobblerProxy.update every 1 second when I run PlexTraktSync with the watch command since version v0.28.x

### Error trace / logs

```text
INFO     Listening for events!
INFO     Server connected: skywalker (1.32.8.7639-fb6452ebf)
INFO     Starting AlertListener:
INFO     Websocket connected
ERROR    KeyError was raised: 'sessionKey'
ERROR    KeyError was raised: 'sessionKey'
INFO     on_play: <Movie:1195:Napoleon>: 0.067816%, State: playing, Watched:
         False, LastViewed: None
INFO     on_play: <Movie:1195:Napoleon>: 0.080719%, State: playing, Watched:
         False, LastViewed: None
WARNING  Rate Limit Exceeded for plextraktsync.trakt.ScrobblerProxy.update(),
         retrying after 1 seconds (try: 1/2)
INFO     on_play: <Movie:1195:Napoleon>: 0.091628%, State: playing, Watched:
         False, LastViewed: None
INFO     on_play: <Movie:1195:Napoleon>: 0.102349%, State: playing, Watched:
         False, LastViewed: None
WARNING  Rate Limit Exceeded for plextraktsync.trakt.ScrobblerProxy.update(),
         retrying after 1 seconds (try: 1/2)
INFO     on_play: <Movie:1195:Napoleon>: 0.113143%, State: playing, Watched:
         False, LastViewed: None
WARNING  Rate Limit Exceeded for plextraktsync.trakt.ScrobblerProxy.update(),
         retrying after 1 seconds (try: 1/2)
```

### Expected behavior

To scrobble without hitting the Rate Limit.

### Steps to reproduce the behavior

1. removed all plextraktsync dockers and configs
2. deployed plextraktsync and login
3. sync works fine, but almost immediately notice Rate Limit Exceeded in the logs.

### Inspect of problematic items

[...]

PR #1757, opened 2024-01-20
Refactor: Add TraktScrobbleWorker worker

glensc:
Moves trakt scrobble api calls to worker. This offloads synchronous requests to background task, making main watch command more responsive. 

It will also aggregate results to update same media object, to submit only latest update for same id: 

```
trakt_id=1, percent=1.11
trakt_id=1, percent=1.23
trakt_id=1, percent=1.40
trakt_id=2, percent=19
```

will become:
```
trakt_id=1, percent=1.40
trakt_id=2, percent=19
```

additionally worker methods respect `@time_limit()`, so methods are not called more often than 1.1 seconds:
- https://github.com/Taxel/PlexTraktSync/blob/ef581d1c05c9f851692b448ada74c1f21e2abe21/plextraktsync/config/__init__.py#L9

### DIFF
--- a/plextraktsync/queue/Queue.py
+++ b/plextraktsync/queue/Queue.py
@@ -29,6 +29,15 @@ class Queue:
     def add_to_history(self, data):
         self.add_queue("add_to_history", data)
 
+    def scrobble_update(self, data):
+        self.add_queue("scrobble_update", data)
+
+    def scrobble_pause(self, data):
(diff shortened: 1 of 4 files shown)


glensc:
Rebased after #1758
> Error: Error running watch command: No module named 'plextraktsync.media'

- https://github.com/Taxel/PlexTraktSync/issues/1724#issuecomment-1902784906

Issue #1788, opened 2024-02-05
Repeatedly Downloading Trakt List, Nothing Else Happens


mocharilla:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists
- [X] I have tried downgrading to find version that can be used as a workaround

### The problem

Sync was working perfectly fine until I updated in the last few days. Now, instead of updating anything, it just keeps repeatedly saying "INFO Downloaded Trakt list" with a set of playlists that repeat over and over. I tried unliking those playlists to see if they specifically were the problem, but it just gave me the errors with different playlists. At no point does it ever do anything else until it eventually gives a completed sync message. It does not give an error at any point.

<img width="850" alt="Screenshot 2024-02-04 at 8 53 31 PM" src="https://github.com/Taxel/PlexTraktSync/assets/158978688/06b34a03-ef27-48d7-9e4d-9223a16be961">

 
<img width="1397" alt="Screenshot 2024-02-04 at 8 56 22 PM" src="https://github.com/Taxel/PlexTraktSync/assets/158978688/79cbd1d7-7c40-44ea-8be7-6ef5f5dcdbd7">

### Steps to reproduce the behavior

1. plextraktsync sync
This persists despite computer restarts, reinstalls of plextraktsync, and downgrading to the last working version.

### Expected behavior

Syncing of the watched status between Trakt and Plex. I added new files since the last sync which are marked as watched in Trakt and properly formatted on the server. This behavior has worked perfectly fine before I updated most recently.

### Workarounds

I tried downgrading, but it did not fix the issue. The version that was working was 27.10 before updating, but the issue persists when downgrading back to that.

### Install method

pipx

### Config file contents

<details>

```yaml
cache:
  path: $PTS_CACHE_DIR/trakt_cache

excluded-libraries:
  - Disney
  - Broadway

config:
  dotenv_override: true

plex:
  timeout: 30

logging:
  append: true
  # Whether to show timestamps in console messages
  console_time: false
  debug: false
  filename: plextraktsync.log
  # Additional logger names to apply filtering
  filter_loggers:
#    - plexapi
#    - requests_cache.backends
#    - requests_cache.backends.base
#    - requests_cache.backends.sqlite
#    - requests_cache.policy.actions
#    - requests_cache.session
#    - trakt.core
#    - urllib3.connectionpool
  filter:
[...]

PR #1791, opened 2024-02-05
Fix TraktUserList load check

glensc:
The self._items would evaluate False if it is an empty array:
- https://github.com/Taxel/PlexTraktSync/issues/1788


glensc:
annoying. no reporter ever tests code from pr even if asked.

Issue #1785, opened 2024-02-02
Error during ratings sync: 'NoneType' object has no attribute 'plex'


JonzaUK:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists
- [X] I have tried downgrading to find version that can be used as a workaround

### The problem

Container stops running.

### Steps to reproduce the behavior

1. Run the container
2. Wait for it to start processing series.

### Error trace / logs

```text
https://gist.github.com/JonzaUK/98c06f229a82ffb35ffe22d6477d122d
```

### Expected behavior

App to finish syncing of libraries

### Install method

Unraid installation (not supported)

### Config file contents

<details>

```yaml
cache:
  path: $PTS_CACHE_DIR/trakt_cache

excluded-libraries:
  - Private
  - Family Holidays

config:
  dotenv_override: true

plex:
  timeout: 30

logging:
  append: true
  # Whether to show timestamps in console messages
  console_time: false
  debug: false
  filename: plextraktsync.log
  # Additional logger names to apply filtering
  filter_loggers:
#    - plexapi
#    - requests_cache.backends
#    - requests_cache.backends.base
#    - requests_cache.backends.sqlite
#    - requests_cache.policy.actions
#    - requests_cache.session
#    - trakt.core
#    - urllib3.connectionpool
  filter:
#    # Filter out all messages with level WARNING
#    - level: WARNING
#    # Filter out message with level WARNING and containing a text
#    - level: WARNING
#      message: "not found on Trakt"
#    - message: "because provider local has no external Id"
#    - message: "because provider none has no external Id"
#    - message: "Retry using search for specific Plex Episode"
#    # Filter out messages by requests_cache
#    - name: requests_cache.backends
#    - name: requests_cache.backends.base
#    - name: requests_cache.backends.sqlite
#    - name: requests_cache.policy.actions
#    - name: requests_cache.session

# settings for 'sync' command (default)
sync:
  plex_to_trakt:
    collection: true
    # Clear collected state of items not present in Plex
    clear_collected: false
    ratings: true
    watched_status: true
    # If plex_to_trakt watchlist=false and trakt_to_plex watchlist=true
    # the Plex watchlist will be overwritten by Trakt watchlist
    watchlist: true
  trakt_to_plex:
    liked_lists: false
    ratings: false
    watched_status: false
[...]

PR #1792, opened 2024-02-06
Fix: Raise error when plex_rating needs show but it is missing

glensc:
Improve logging for https://github.com/Taxel/PlexTraktSync/issues/1785


glensc:
annoying. no reporter ever tests code from pr even if asked.

Issue #1808, opened 2024-02-19
Feature request: Config per library


SebbeJohansson:
### Feature description
A way to have different settings per library as to allow for syncing only specific things in specific libraries.

### Use case

I want to be able to only sync the collection status from plex to trakt in one specific library, but sync everything for all other libraries.

### Workarounds

I guess a way to do this is to have two separate docker-compose containers, but that's not ideal for me.

### Plan to implement

Not currently.

glensc:
this is already there, just nobody wants to document anything:
- https://github.com/Taxel/PlexTraktSync/pull/1569

it has "docs needed" label

glensc:
> I guess a way to do this is to have two separate docker-compose containers, but that's not ideal for me.

you could also have two different Servers in same servers config

SebbeJohansson:
@glensc heh... I should have figured that it was already a thing. I will look into documenting it. Closing.

glensc:
Use `--server=xxx` to set what server to use. No need to login or anything.

SebbeJohansson:
Hmmm okay I'll give it a try. Thanks!

SebbeJohansson:
I wasnt sure on how its recommended to use ofilia with this kind of a server, but I managed to get it to work atleast.

PR #1814, opened 2024-02-20
Config added: Per server/library configuration example and instructions.

SebbeJohansson:
- Adds instructions on how to setup configuration per server and per library.
- Adds examples of how to setup configuration per server and per library.
- Adds instructions and examples of how to schedule with ofilia if multiple servers are used.

Fixes https://github.com/Taxel/PlexTraktSync/issues/1808

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,7 +427,9 @@ If you want to run these jobs using `ofelia`, you can do so by running something
 ```
 If you are running only one PlexTraktSync container, you need to make sure that the two jobs Ofelia jobs don't run at the same time. Depending on how long it takes for the job to run on your server you might have to keep the schedule of the two jobs seperated with a few minutes or a few hours. If you have two different PlexTraktSync containers in your docker compose, you can run them at the same time.
 
-The above config means that a job is running every 6 hours, alternating between the two "servers". The PlexTraktSync container also has a profile (docker-compose feature) called "schedule" which means that it won't run automatically when you run for example `docker-compose up`.
+The above config means that a job is running every 6 hours, alternating between the two "servers". The PlexTraktSync container also has a [docker compose profile] called "schedule" which means that it won't run automatically when you run for example `docker-compose up`.
+
+[docker compose profile]: https://docs.docker.com/compose/profiles/
 
 ### Logging
 
--- a/README.md
+++ b/README.md
@@ -396,7 +396,7 @@ This can also be used to have different configs between different libraries. To 
 The above config would make it so that the "Movies" library syncs both ratings and watched status, while the "TV Shows" library only syncs ratings.
 To then run the sync you need to specify `--server Example1` or `--server Example2` to run the sync for that specific server. Running the sync command without it will result in running the sync using the default config.
 
-If you want to run these jobs using `oflia`, you can do so by running something similar to this in your `docker-compose.yml`:
+If you want to run these jobs using `ofelia`, you can do so by running something similar to this in your `docker-compose.yml`:
 ```yml
   plextraktsync:
     image: ghcr.io/taxel/plextraktsync
--- a/README.md
+++ b/README.md
@@ -361,7 +361,7 @@ If you want to specify your config per server you can do so inside of `servers.y
 ```
 Using `sync` in a server config overrides the global sync-config in `config.yml`.
 
-This can also be used to have different configs between different libraries. To be able to do this you specifiy the number of servers you need (most likely equal to the number of different config setups you need). For example:
+This can also be used to have different configs between different libraries. To be able to do this you specify the number of servers you need (most likely equal to the number of different config setups you need). For example:
 
 ```yml
   Example1:
--- a/README.md
+++ b/README.md
@@ -359,7 +359,7 @@ If you want to specify your config per server you can do so inside of `servers.y
         trakt_to_plex:
           liked_lists: false
 ```
-Only `sync` can be used inside of the config key-value. Keep in mind that the sync-config set in `servers.yml` overrides the global sync-config in `config.yml`.
+Using `sync` in a server config overrides the global sync-config in `config.yml`.
 
 This can also be used to have different configs between different libraries. To be able to do this you specifiy the number of servers you need (most likely equal to the number of different config setups you need). For example:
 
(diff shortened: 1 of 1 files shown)


SebbeJohansson:
As discussed in: #1808

README.md
```diff
@@ -343,6 +343,90 @@ $ plextraktsync --server=Example1 info
...
+    urls:
+      - http://localhost:32400
+    config:
+      sync:
+        plex_to_trakt:
+          collection: true
+        trakt_to_plex:
+          liked_lists: false
+```
+Both `watch` and `sync` can be used inside of the config key-value.
```
glensc, line 362: no. only sync is supported. where you took watch is supported too?

README.md
```diff
@@ -343,6 +343,90 @@ $ plextraktsync --server=Example1 info
...
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+    labels:
+      ofelia.job-run.plextraktsync.schedule: "0 6,18 * * *"
+      ofelia.job-run.plextraktsync.container: "plextraktsync"
+      ofelia.job-run.plextraktsync.command: "--server 'Example1' sync"
+      ofelia.job-run.plextraktsync2.schedule: "0 12,0 * * *"
+      ofelia.job-run.plextraktsync2.container: "plextraktsync"
+      ofelia.job-run.plextraktsync2.command: "--server 'Example2' sync"
+```
+This means that a job is running every 6 hours, alternating between the two "servers". Keep in mind that since this example assumes that you are only running one container, you need to make sure that the jobs don't overlap.
```
glensc, line 428: why jobs can't overlap?

glensc:
general note: it should be noted that the server configs override global config.yml

README.md
SebbeJohansson: I think read it somewhere, but can't find that specific info I read. It could be that I misunderstood #1206 thinking it meant that watch was supported. I'll update it later today.

README.md
SebbeJohansson: Because if the job takes a long time (say over half a second) Ofelia doesn't want to create the same container again. Maybe there is a cleaner way? Not super used to Ofelia.

SebbeJohansson:
> general note: it should be noted that the server configs override global config.yml

@glensc good point. I'll note that somewhere.

README.md
glensc: so, that just means it runs only one job at a time.

either remove the warning, or explain what it's for.

README.md
```diff
@@ -343,6 +343,90 @@ $ plextraktsync --server=Example1 info
...
+    urls:
+      - http://localhost:32400
+    config:
+      sync:
+        plex_to_trakt:
+          collection: true
+        trakt_to_plex:
+          liked_lists: false
+```
+Only `sync` can be used inside of the config key-value. Keep in mind that the sync-config set in `servers.yml` overrides the global sync-config in `config.yml`.
```
glensc, line 362: "only sync" is not true either, because other section describes "library" and "excluded-libraries".

perhaps reword as: "sync" section overrides can be used or something like that.

SebbeJohansson: That is very true. Nice catch. What do you think about: "Using `sync` in a server config overrides the global sync-config in `config.yml`." as an alternative?

SebbeJohansson: Commited an alternative! Let me know.

README.md
```diff
@@ -343,6 +343,92 @@ $ plextraktsync --server=Example1 info
...
+    config:
+      sync:
+        plex_to_trakt:
+          collection: true
+        trakt_to_plex:
+          liked_lists: false
+```
+Using `sync` in a server config overrides the global sync-config in `config.yml`.
+
+This can also be used to have different configs between different libraries. To be able to do this you specifiy the number of servers you need (most likely equal to the number of different config setups you need). For example:
```
glensc, line 364:
```suggestion
This can also be used to have different configs between different libraries. To be able to do this you specify the number of servers you need (most likely equal to the number of different config setups you need). For example:
```

README.md
```diff
@@ -343,6 +343,92 @@ $ plextraktsync --server=Example1 info
...
+      sync:
+        plex_to_trakt:
+          ratings: true
+          watched_status: false
+        trakt_to_plex:
+          ratings: true
+          watched_status: false
+```
+The above config would make it so that the "Movies" library syncs both ratings and watched status, while the "TV Shows" library only syncs ratings.
+To then run the sync you need to specify `--server Example1` or `--server Example2` to run the sync for that specific server. Running the sync command without it will result in running the sync using the default config.
```
glensc, line 397:
```suggestion
To then run the sync you need to specify `--server Example1` or `--server Example2` to run the sync for that specific server.

Running the sync command without `--server` will use default server from `.env`
```

again. not true statement detected. --server doesn't specify what configs to override, but rather what server to pick. your default server could be either of those examples or something completely different.

README.md
```diff
@@ -343,6 +343,92 @@ $ plextraktsync --server=Example1 info
...
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+    labels:
+      ofelia.job-run.plextraktsync.schedule: "0 6,18 * * *"
+      ofelia.job-run.plextraktsync.container: "plextraktsync"
+      ofelia.job-run.plextraktsync.command: "--server 'Example1' sync"
+      ofelia.job-run.plextraktsync2.schedule: "0 12,0 * * *"
+      ofelia.job-run.plextraktsync2.container: "plextraktsync"
+      ofelia.job-run.plextraktsync2.command: "--server 'Example2' sync"
+```
+If you are running only one PlexTraktSync container, you need to make sure that the two jobs Ofelia jobs don't run at the same time. Depending on how long it takes for the job to run on your server you might have to keep the schedule of the two jobs seperated with a few minutes or a few hours. If you have two different PlexTraktSync containers in your docker compose, you can run them at the same time.
```
glensc, line 428:
```suggestion
If you are running only one PlexTraktSync container, you need to make sure that the two jobs Ofelia jobs don't run at the same time. Ofelia skips scheduling a new job run if the previous job is still running.

  Depending on how long it takes for the job to run on your server you might have to keep the schedule of the two jobs seperated with a few minutes or a few hours. If you have two different PlexTraktSync containers in your docker compose, you can run them at the same time.
```

you still failed to describe the reasoning why overlapping must be avoided.

README.md
```diff
@@ -343,6 +343,92 @@ $ plextraktsync --server=Example1 info
...
+
+The above config means that a job is running every 6 hours, alternating between the two "servers". The PlexTraktSync container also has a profile (docker-compose feature) called "schedule" which means that it won't run automatically when you run for example `docker-compose up`.
```
glensc, line 430:
```suggestion
The above config means that a job is running every 6 hours, alternating between the two "servers". The PlexTraktSync container also has a [docker compose profile] called "schedule" which means that it won't run automatically when you run for example `docker-compose up`.

[docker compose profile]: https://docs.docker.com/compose/profiles/
```

add link to doc

SebbeJohansson: Good idea!

README.md
SebbeJohansson: lol true.

README.md
SebbeJohansson: Yea! sorry.
Didnt know that default server were saved somewhere.

README.md
SebbeJohansson: LET ME TELL YOu. I misspelled specify so many times last night... I guess you caught one of the ones I missed.

PR #1892, opened 2024-04-08
Refactor: Use pluggy framework to extract Sync to plugins

glensc:
Why: The logic in Sync has be come complicated.

Extra: Possible runtime speedup:
Using plugins can skip checks of every media by if plugin for some feature not enabled at all.

NOTE: `WatchListPlugin` is still too complicated: contains two watchlist sync logics

- https://pluggy.readthedocs.io/en/latest/#a-toy-example


glensc:
side-effect:
1.  Pugin all class methods are resolved:  `method: object = getattr(plugin, name)`, this results all @cached_property being resolved at `register()` time.

bug:
1. plugin methods declared with `@hookimpl`, using default values makes those values being "invisible": they never get assigned a value, their default value is always used.

Issue #1949, opened 2024-05-16
Sync and unmatched error on 0.30.7


AndreiArdelean1:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if identical issue already exists
- [X] I have tried downgrading to find version that can be used as a workaround

### The problem

Running the following commands fails when running version 0.30.7:

`python3 -m plextraktsync --no-progressbar unmatched`
`python3 -m plextraktsync --no-progressbar sync --sync=shows`
`python3 -m plextraktsync --no-progressbar sync --sync=movies`

### Steps to reproduce the behavior

Have version 0.30.7 installed in docker

Run commands:
`python3 -m plextraktsync --no-progressbar unmatched`
`python3 -m plextraktsync --no-progressbar sync --sync=shows`
`python3 -m plextraktsync --no-progressbar sync --sync=movies`

### Error trace / logs

```text
`python3 -m plextraktsync --no-progressbar unmatched`:

         ╭───────────────── Traceback (most recent call last) ─────────────────╮
         │ /app/plextraktsync/cli.py:26 in wrap                                │
         │                                                                     │
         │    23 │   │   │   cmd = getattr(module, name)                       │
         │    24 │   │   │                                                     │
         │    25 │   │   │   try:                                              │
         │ ❱  26 │   │   │   │   cmd(*args, **kwargs)                          │
         │    27 │   │   │   except EOFError as e:                             │
         │    28 │   │   │   │   raise ClickException(f"Program requested term │
         │       connected: {e}")                                              │
         │    29 │   │   │   except ClickException as e:                       │
         │                                                                     │
         │ /app/plextraktsync/commands/unmatched.py:23 in unmatched            │
         │                                                                     │
         │   20 │   │   │   if all(guid.local for guid in pm.guids):           │
         │   21 │   │   │   │   failed.append(pm)                              │
         │   22 │   else:                                                      │
         │ ❱ 23 │   │   for pm in walker.get_plex_movies():                    │
         │   24 │   │   │   movie = mf.resolve_any(pm)                         │
         │   25 │   │   │   if not movie:                                      │
         │   26 │   │   │   │   failed.append(pm)                              │
         ╰─────────────────────────────────────────────────────────────────────╯
         TypeError: 'async_generator' object is not iterable                    
Error: Error running unmatched command: 'async_generator' object is not iterable



[...]

PR #1950, opened 2024-05-16
Fix: No progressbar launches sync code

glensc:
Fixes https://github.com/Taxel/PlexTraktSync/issues/1949


glensc:
Tried to create wrapper, but `aiter()`, `anext()` needs python 3.10:
- https://github.com/python/cpython/issues/76042

PR #1994, opened 2024-06-18
Update README.md

Geezus42:
`docker-compose` --> `docker compose`

The name of the command has changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ Note: `main` is development version and reporting bugs against development versi
 To run sync:
 
 ```
-docker-compose run --rm plextraktsync sync
+docker compose run --rm plextraktsync sync
 ```
 
 The container will stop after the sync is completed. Read Setup section to run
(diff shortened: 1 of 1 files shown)


glensc:
1. clickbait. pr title should describe the changes
2. both commands should be supported. so this is no-go

Geezus42:
Both commands are not supported unless you have aliases. If you run `docker-compose` on a system without aliases, it will tell you the command does not exist. I understand some distros/installers may include the alias by default, but it is stop gap that should be phased out since the command syntax has been changed.

Geezus42:
Both commands are not supported unless you have aliases. If you run `docker-compose` on a system without aliases, it will tell you the command does not exist. I understand some distros/installers may include the alias by default, but it is stop gap that should be phased out since the command syntax has been changed.

Geezus42:
There fixed the title.

PR #1998, opened 2024-06-19
Update README.md: change docker-compose to docker compose because command syntax has changed

Geezus42:
`docker-compose` --> `docker compose`

The official syntax of the command has changed. Many distributions and installers will automatically create an alias, but this should no be relied upon since it will not always be there. Documentation should be updated to reflect the standard syntax as defined by Docker. 

```
$ docker-compose
The program 'docker-compose' is currently not installed. It is provided by
several packages. You can install it by typing one of the following:

$ docker compose
Usage:  docker compose [OPTIONS] COMMAND
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ Note: `main` is development version and reporting bugs against development versi
 To run sync:
 
 ```
-docker-compose run --rm plextraktsync sync
+docker compose run --rm plextraktsync sync
 ```
 
 The container will stop after the sync is completed. Read Setup section to run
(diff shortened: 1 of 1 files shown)


glensc:
as written in https://github.com/Taxel/PlexTraktSync/pull/1995:
- both should be supported.

Geezus42:
I don't understand the problem. You just referenced your own comment, which I already read, and did not provide any additional information. Docker has not said both will be supported. In fact, they ended support for Compose V1 in 2021. 

The syntax of the command has officially been changed. If the old syntax works on your system, then it is a crutch that will eventually go away. It is better to simply learn the new syntax and update documentation to accordingly. Again, I do not understand the push back on such a minor change. 

[Can I still use Compose V1 if I want to?](https://docs.docker.com/compose/migrate/#can-i-still-use-compose-v1-if-i-want-to)
image

PR #2043, opened 2024-08-31
Feature: Add update-playlist command: Remove watched items

glensc:
Add `update-playlist` subcommand. Currently only able to remove watched (local and remote) items from playlist.


glensc:
NOTE: Broken: It removed all remote playlist items on save.

PR #2082, opened 2024-10-20
CI: Setup renovate bot

glensc:
GitHub Dependabot is quite dumb, problems I've reported to them haven't been fixed for years. Perhaps renovate does better, or is more open actually fixing things.

Based on the file found on the internet:
- https://github.com/prskr/goveal/blob/90299a1405e95673bed349eb2847b918b014c73c/.gitea/workflows/renovate.yaml#L2

Setup renovate on github CI, because using service requires repo permissions I do not have:
- https://docs.renovatebot.com/getting-started/installing-onboarding/

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -1,0 +1,21 @@
+name: Renovate
+on:
+  schedule:
+    # The "*" (#42, asterisk) character has special semantics in YAML, so this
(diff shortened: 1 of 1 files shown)


glensc:
Would still need to setup github token:
- https://github.com/renovatebot/github-action#token

..., which I don't have access to...

PR #2087, opened 2024-10-21
Cleanup: Remove all nonsensical python_version ranges

glensc:
The application needs >= 3.9 anyway, so having them for each dep is superfluous

Basically, we need just things like these:

> `exceptiongroup==1.2.2; python_version < '3.11'`


glensc:
Dependabot will likely add them back one by one, but at least it's cleaned from previous bugs for now.

PR #2108, opened 2024-12-11
added info to documentation on creating ciritcal files without access to commandline

kidforpm:
added a little help to hint at creating the necessary files elsewhere when docker image cannot be run from command-line (like on Synology NAS).


glensc:
Update PR title to summarize changes

Issue #2139, opened 2025-01-05
Cant schedule using ofelia because of "No such container: plextraktsync"


SebbeJohansson:
### Confirmation

- [X] I have read the [README.md](https://github.com/Taxel/PlexTraktSync#readme) on the project homepage
- [X] I have checked if an identical issue or discussion already exists
- [X] I have tried downgrading to find version that can be used as a workaround

### The problem

Running ofelia with the suggested settings and setup is not working.
```
administrator@big-vault:~$ sudo docker-compose logs scheduler
2025-01-05T00:18:00.017Z  common.go:121 ▶ ERROR [Job "plextraktsync" (df74273cad1b)] Finished in "963.091µs", failed: true, skipped: false, error: No such container: plextraktsync
```

### Steps to reproduce the behavior

1. Setup PlexTraktSync according to the documentation.
2. Setup ofelia according to the documentation.
3. Start the containers using `sudo docker-compose up -d` or `sudo docker compose up -d`.
4. Wait for the schedule to run.
5. Check the logs.

### Error trace / logs

```text
administrator@big-vault:~$ sudo docker-compose logs scheduler
2025-01-05T00:18:00.016Z  common.go:125 ▶ NOTICE [Job "plextraktsync" (df74273cad1b)] Started - --server 'The Big Vault Movies' sync
2025-01-05T00:18:00.017Z  common.go:121 ▶ ERROR [Job "plextraktsync" (df74273cad1b)] Finished in "963.091µs", failed: true, skipped: false, error: No such container: plextraktsync
```

### Expected behavior

I expect the scheduled job to run.

### Config file contents

```yaml
plextraktsync:
    image: ghcr.io/taxel/plextraktsync
    command: sync
    container_name: plextraktsync
    profiles: ["schedule"]
    volumes:
      - /configs/mediarr/plextraktsync:/app/config
    environment:
      - PUID=1000
      - PGID=1000
    depends_on:
      - plex
  scheduler:
    image: mcuadros/ofelia:latest
    container_name: scheduler
    command: daemon --docker
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
    labels:
      ofelia.job-run.plextraktsync.schedule: "0 6,18 * * *"
      ofelia.job-run.plextraktsync.container: "plextraktsync"
      ofelia.job-run.plextraktsync.command: "--server 'The Big Vault Movies' sync"
[...]

PR #2142, opened 2025-01-07
Config updated: Working per server/library scheduling with ofelia

SebbeJohansson:
- Removes "schedule" profile in multi server setup since it doesnt work.
- Updates image versions so that it is cohesive through the whole documentation (latest on all).
- Updates config volume path so that it is the same in multi server config as it is in the single server config.

After discussions and testing for #2139.

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,7 +455,6 @@ services:
     image: ghcr.io/taxel/plextraktsync
     command: sync
     container_name: plextraktsync
-    profiles: ["schedule"]
     volumes:
       - /configs/mediarr/plextraktsync:/app/config
     environment:
@@ -464,7 +463,7 @@ services:
     depends_on:
       - plex
   scheduler:
-    image: mcuadros/ofelia:latest
+    image: mcuadros/ofelia
     container_name: scheduler
     command: daemon --docker
     restart: unless-stopped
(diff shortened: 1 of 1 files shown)


simonc56:
We have no prove that "schedule" profile doesn not work, I suggest you remove this line in first post.
I agree to remove the `profile` stuff which is not needed and misleading users unable to read the doc.

simonc56 (approved):

SebbeJohansson:
@simonc56 The proof would be in that it doesn't work anymore on a clean system. Feel free to test it yourself.

README.md
```diff
@@ -452,12 +452,11 @@ something similar to this in your `docker-compose.yml`:
 ```yml
 services:
   plextraktsync:
-    image: ghcr.io/taxel/plextraktsync
+    image: ghcr.io/taxel/plextraktsync:latest
```
glensc, line 455: remove this chunk

glensc:
Ideally, I'd see one change per pull request. lower chances that unrelated changes don't block whole pr merge. you made three changes, one of them is rejected.

SebbeJohansson:
@glensc Just to be clear, you are against adding ":latest" to the places where it is missing? Should we remove it from all places? I can revert the chunks here and make a new pr for how ever you want it.

SebbeJohansson:
@glensc okay so the documentation should have different images for different parts. Im not sure I'm a fan of that, but I'm not gonna fight it.

I have reverted the chunks you specified and added versions to the ofelia image.

glensc:
@SebbeJohansson, as I said, ofelia should be pinned to some major version. so if they release something breaking, the docs here would still be acurate. As for plextraktsync, it's current repository, it's latest will be supported, rather than any other major.

PR #2143, opened 2025-01-08
Feature: Add l.plex.tv and watch.plex.tv url support to inspect

glensc:


glensc:
How to resolve urls like https://watch.plex.tv/movie/heavier-trip to plex hash?

glensc:
Seems the `utm_content` contains plex_id that can be used in discover

PR #2383, opened 2025-12-28
Log resolved logging configuration at startup

bluemeaford:
This adds a small startup info log that prints the resolved log file path and effective logging settings (append/rotation).

It helps diagnose Docker/mount-related issues where the app may be reading or writing logs from a different location than expected. No behaviour changes.

### Example output

Example startup log line (paths may vary by environment):

```text
INFO plextraktsync.logger: Resolved logging config:
log_file=/app/config/plextraktsync.log append=False rotation=False


glensc:
Put example output to PR description.

PR #2473, opened 2026-05-16
perf: reduce redundant Trakt API calls during sync

Gilmoursa:
Three related fixes to reduce how many Trakt API calls are made during a sync, addressing rate limit errors on large libraries.

### 1. `search_by_id()`: add `@time_limit()` throttle (proactive)

`search_by_id()` had `@rate_limit()` (reactive — waits `retry-after` seconds once the limit is already hit) but was missing `@time_limit()` (proactive — enforces a minimum delay between calls). During a full library sync the function fires one unthrottled call per Plex item, rapidly exhausting the rate limit. This shares the existing `Timer(TRAKT_POST_DELAY)` instance with POST operations, capping all outbound Trakt requests to one per 1.1s. All other Trakt methods already had both decorators; `search_by_id()` was the only exception.

### 2. `TraktApi.show_collection`: `@property` → `@cached_property`

Every other collection (`movie_collection`, `watched_movies`, `collected_shows`) was already `@cached_property`. `show_collection` was the sole exception, triggering a fresh `me.show_collection` API call each time it was accessed. `episodes_collection` is already cached but depends on `show_collection` as its source, so this eliminates redundant fetches (e.g. in `ClearCollectedPlugin`).

### 3. `Walker.get_plex_episodes()`: cache show lookups by `grandparentGuid`

When syncing specific episodes (via `plan.episodes`), the function called `resolve_guid()` — and thus `search_by_id()` — for every episode's `grandparentGuid` with no deduplication. Syncing 200 episodes across 3 shows triggered 200 API calls instead of 3. The section-based path in `find_episodes()` already used a `show_cache` dict for exactly this; `get_plex_episodes()` was missing it.

> Note: I used AI-assisted coding for these changes.

### DIFF
--- a/plextraktsync/config/__init__.py
+++ b/plextraktsync/config/__init__.py
@@ -8,8 +8,15 @@ PLEX_PLATFORM = "PlexTraktSync"
 
 """
 Constant in seconds for how much to wait between Trakt POST API calls.
+POST requests are limited to 1 call/second per user.
 """
 TRAKT_POST_DELAY = 1.1
+
+"""
+Constant in seconds for how much to wait between Trakt GET API calls.
+GET requests are limited to 1,000 calls per 5 minutes (~0.3s between calls).
+"""
+TRAKT_GET_DELAY = 0.3
 
 """
 Constants in seconds for the margin added to retry-after delay to account for network jitter in rate limiting retries.
--- a/plextraktsync/decorators/get_limit.py
+++ b/plextraktsync/decorators/get_limit.py
@@ -1,0 +1,20 @@
+from __future__ import annotations
+
+from decorator import decorator
+
+from plextraktsync.config import TRAKT_GET_DELAY
+from plextraktsync.util.Timer import Timer
+
+timer = Timer(TRAKT_GET_DELAY)
+
+
+@decorator
+def get_limit(fn, *args, **kwargs):
+    """
+    Throttles GET calls not to exceed Trakt's GET rate limit (1,000 per 5 minutes).
+    Uses a separate timer from time_limit to avoid conflating GET and POST budgets.
+    """
+
+    timer.wait_if_needed()
+
+    return fn(*args, **kwargs)
--- a/plextraktsync/plan/Walker.py
+++ b/plextraktsync/plan/Walker.py
@@ -150,10 +150,14 @@ class Walker(SetWindowTitle):
             yield show
 
     async def get_plex_episodes(self, episodes: list[Episode]) -> Generator[Media, Any, None]:
+        show_cache: dict[str, Media | None] = {}
         it = self.progressbar(episodes, desc="Processing episodes")
         async for pe in it:
-            guid = PlexGuid(pe.grandparentGuid, "show")
-            show = self.mf.resolve_guid(guid)
+            grandparent_guid = pe.grandparentGuid
+            if grandparent_guid not in show_cache:
+                guid = PlexGuid(grandparent_guid, "show")
+                show_cache[grandparent_guid] = self.mf.resolve_guid(guid)
+            show = show_cache[grandparent_guid]
             if not show:
                 continue
             me = self.mf.resolve_any(PlexLibraryItem(pe, plex=self.plex), show)
--- a/plextraktsync/trakt/TraktApi.py
+++ b/plextraktsync/trakt/TraktApi.py
@@ -101,7 +101,7 @@ class TraktApi:
     def movie_collection(self):
         return self.me.movie_collection
 
-    @property
+    @cached_property
     @rate_limit()
     @retry()
     def show_collection(self):
--- a/plextraktsync/trakt/TraktApi.py
+++ b/plextraktsync/trakt/TraktApi.py
@@ -18,6 +18,7 @@ from trakt.errors import (
 
 from plextraktsync import pytrakt_extensions
 from plextraktsync.decorators.flatten import flatten_list
+from plextraktsync.decorators.get_limit import get_limit
 from plextraktsync.decorators.rate_limit import rate_limit
 from plextraktsync.decorators.retry import retry
 from plextraktsync.decorators.time_limit import time_limit
@@ -282,7 +283,7 @@ class TraktApi:
             raise RuntimeError(f"find_by_slug: Unable to find with slug: {slug}")
 
     @rate_limit()
-    @time_limit()
+    @get_limit()
     def search_by_id(self, media_id: str, id_type: str, media_type: str) -> TVShow | Movie | None:
         if id_type == "tvdb" and media_type == "movie":
             # Skip invalid search.
--- a/plextraktsync/trakt/TraktApi.py
+++ b/plextraktsync/trakt/TraktApi.py
@@ -282,6 +282,7 @@ class TraktApi:
             raise RuntimeError(f"find_by_slug: Unable to find with slug: {slug}")
 
     @rate_limit()
+    @time_limit()
     def search_by_id(self, media_id: str, id_type: str, media_type: str) -> TVShow | Movie | None:
         if id_type == "tvdb" and media_type == "movie":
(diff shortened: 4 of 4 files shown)


simonc56:
`search_by_id()` triggers a GET request.
The [limit](https://trakt.docs.apiary.io/#introduction/rate-limiting) for GET requests is 1000 calls every 5 minutes.
The 1 call per second is for POST requests => 300 calls for 5 minutes.
Why slowing to 300 calls/5min instead of the 1k calls/5min allowed ?
Did you conduct a sync test before/after to measure the improvement ?

Gilmoursa:
Good catch — that's a valid distinction I missed. GET and POST have separate budgets on Trakt, and conflating them was too conservative.

Updated the fix: added a dedicated `TRAKT_GET_DELAY = 0.3s` constant and a separate `get_limit` decorator with its own timer, then applied `@get_limit()` to `search_by_id()` in place of `@time_limit()`. That keeps GET requests at ~1,000/5min (matching Trakt's actual GET limit) without touching the POST timer.

On the test question — no benchmark data, sorry. The original fix was based on the reported symptom (repeated "Rate Limit Exceeded" warnings with 300s retries on `search_by_id` during a full library sync) rather than measured throughput. If you have a large library to test against, I'd welcome a before/after comparison.

simonc56:
What problem are you trying to fix with this new 0.3s throttle ?

> repeated "Rate Limit Exceeded" warnings with 300s retries on search_by_id during a full library sync

Explain why this is a problem.

Gilmoursa:
Fair point, and thanks for pushing back on this.

During a full library sync on a large collection, I was seeing repeated `Rate Limit Exceeded` responses on `search_by_id()` calls, each triggering a 300-second retry wait. Over the course of a full sync that added up to significant delays. My instinct was to proactively throttle the calls before hitting the limit rather than react after the fact.

What I missed — and you're right to call it out — is that `search_by_id()` is a GET request, so the applicable limit is 1,000 calls per 5 minutes, not 300. Adding `@time_limit()` would cap it at the POST rate unnecessarily. I was pattern-matching to the other throttled methods without checking whether they shared the same HTTP method and rate tier.

So the real question becomes: why was a GET endpoint hitting rate limits at all on a large sync? It's possible the issue is upstream — the `resolve_guid()` deduplication in Walker.py (also in this PR) would reduce `search_by_id()` call volume significantly on its own. I haven't isolated whether that change alone resolves the rate limit warnings, or if there's still a need for throttling calibrated to the actual GET limit.

Happy to remove the `@time_limit()` change and retest with just the `cached_property` and `show_cache` fixes in place to see if the rate limit warnings disappear. Would that be the right next step?

simonc56:
You still don't explain why hitting rate limit is a problem...

>  resolves the rate limit warnings

It's a warning, not an error. What do you want to resolve ?

Gilmoursa:
That's fair — it is a warning and the retry logic does handle it eventually. The motivation was user experience rather than correctness: when `search_by_id()` fires unthrottled on a large library, hitting the limit means waiting the full `retry-after` period (typically 300 seconds) before continuing — potentially multiple times per sync. For users with libraries in the thousands, that turns a sync into an hours-long process punctuated by 5-minute pauses.

That said, you're right that for smaller libraries that never approach the limit, the throttle just adds unnecessary delay. If the consensus is that reactive handling is sufficient and users should simply expect the waits, I'm happy to drop the change entirely — no attachment to it!

simonc56:
The way you write and the em dash in all your comments means you're not only asking AI to code for you but also copy/pasting AI text to talk to me 😞 
Can't you write your own words ?
One would wonder if you even understand what you're copying.

Gilmoursa:
Because of my disability, typing is sometimes difficult for me, so I will use voice dictation software or sometimes AI to help write my thoughts down. The PlexTraktSync tool is a project I am interested in because I use both Trakt and Plex daily. 

It seems I have offended you. I will close this pull request and work on other projects related to my media server. I hope you have a good weekend.

simonc56:
> When syncing specific episodes (via plan.episodes), the function called `resolve_guid()` — and thus `search_by_id()` — for every episode's `grandparentGuid` with no deduplication. Syncing 200 episodes across 3 shows triggered 200 API calls instead of 3

Do you have logs of this ?
Because all `search_by_id` trakt requests have 1-day cache so it shouldn't happen.

Gilmoursa:
I wanted to share some logs from my Unraid server that show the rate limit hits are real and happening in production. During a collection sync on May 16, `search_by_id` hit the rate limit 3 times within 12 minutes, each triggering a 300 second retry wait:

 ```
 2026-05-16 11:07:41,121 WARNING[plextraktsync.decorators.rate_limit]:Rate Limit Exceeded for plextraktsync.trakt.TraktApi.search_by_id(), retrying after 300 seconds (try: 1/5)
 2026-05-16 11:13:25,344 WARNING[plextraktsync.decorators.rate_limit]:Rate Limit Exceeded for plextraktsync.trakt.TraktApi.search_by_id(), retrying after 300 seconds (try: 1/5)
 2026-05-16 11:19:20,320 WARNING[plextraktsync.decorators.rate_limit]:Rate Limit Exceeded for plextraktsync.trakt.TraktApi.search_by_id(), retrying after 300 seconds (try: 1/5)
 ```

 The surrounding context shows these are firing during `AddCollectionPlugin` processing movies rapidly back to back, so the 1-day cache does not appear to be preventing the issue in this scenario. I understand the approach may not have been the right fix, but wanted to share this in case it is useful for tracking down the root cause. I will not contribute to this project anymore. Have a good weekend.

simonc56:
Thank you for your contribution, have a good weekend too.

JunaidDawood:
@simonc56:

INFO     Sync Movie sections: ['Movies']
INFO     Sync Show sections: ['TV Shows']
INFO     Enable sync plugin 'AddCollectionPlugin': False
INFO     Enable sync plugin 'ClearCollectedPlugin': False
INFO     Enable sync plugin 'LikedListsPlugin': False
INFO     Enable sync plugin 'SyncRatingsPlugin': True
INFO     Enable sync plugin 'SyncWatchedPlugin': True
INFO     Enable sync plugin 'TraktListsPlugin': False
INFO     Enable sync plugin 'WatchListPlugin': True
INFO     Enable sync plugin 'WatchProgressPlugin': False
Processing Movies 100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 913/913  [ 0:01:09 < 0:00:00 , 14 it/s ]
INFO     Movies processed in 1 minute and 10.32 seconds
INFO     Preload shows data
Processing TV Shows 100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 146/146  [ 0:00:00 < 0:00:00 , 1,531 it/s ]
INFO     TV Shows processed in 225 milliseconds and 777 microseconds
INFO     Preloaded shows data (146 shows)
INFO     Preload shows matches
WARNING  Rate Limit Exceeded for plextraktsync.trakt.TraktApi.search_by_id(), retrying after 300 seconds (try: 1/5)
Processing show matches  99% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╺ 144/146  [ 0:02:48 < 0:00:01 , 14 it/s ]

glensc:
👎🏿 for this pr. don't enforce delay un-neccessarily. calls will be blocked by trakt api anyway if you do too many requests. eventually it will take longer than pausing the 300 seconds.

but what you may play with, is to adjust cache rules in your `config.yml`, see project default `config.default.yml` what rules to override. `plextraktsync config --urls-expire-after` to show merged config. enable debug logging to see what other patterns you may want to increase cache.

note the http caching only helps if you re-run scan for existing library. nothing much to do for fresh library without cache.

PR #2478, opened 2026-05-20
feat(healthcheck): add healthcheck command and Docker healthcheck

### LINKED ISSUE #2323: Token refresh does not get picked up by watch mode container if it gets refreshed in sync mode container.
> @simonc56 current problem with the latest release:
> ```
> 2025-09-20T12:19:43.406990490Z ERROR    400 - Unable to refresh expired OAuth token (invalid_grant) The        
> 2025-09-20T12:19:43.407014540Z          provided authorization grant is invalid, expired, revoked, does not    
> 2025-09-20T12:19:43.407018730Z          match the redirection URI used in the authorization request, or was    
> 2025-09-20T12:19:43.407021832Z          issued to another client.                                              
> 2025-09-20T12:19:43.420209919Z ERROR    Max token refresh attempts reached. Manual intervention required. 
> ```
> 
> as soon as I start watching something. It'll just constantly spits out the `Max token refresh attempts reached. Manual intervention required.` and the `Unauthorized - OAuth must be provided` until I restart the container through either my auto restart on unhealthy or manually restarting it.
> 
> I run both sync and watch. Watch, I have running 24/7 (which is where the problem arises) while sync, I have run every hour (doesn't have this problem since the container only starts up long enough to do a sync and then shuts down again until the next hour). It seems that if the sync ends up running the refresh for the token, then the watch container doesn't pick up the new token and ends up running into the above problem.
> 
> PlexTraktSync info:
> ```
> PlexTraktSync Version: 0.34.15
> Python Version: 3.13.7 (main, Aug 15 2025, 22:06:02) [GCC 14.2.0]
> Plex API Version: 4.17.1
> Trakt API Version: 4.2.2
> ``` 

 _Originally posted by @JourneyOver in [#1546](https://github.com/Taxel/PlexTraktSync/issues/1546#issuecomment-3314969364)_

DennisGaida:
`watch` silently fails on error (e.g. OAUTH, see #2323), but would benefit from an error state leveraging docker healthchecks.

This PR adds `plextraktsync healthcheck` to detect unhealthy containers in watch mode. Scans the configured log file for ERROR or CRITICAL level entries within a time window (default 3600s) and exits 1 if any are found, letting Docker mark the container unhealthy after --retries=3.

The log probe tails only the last max_age * 256 bytes so it stays fast on large log files between rotations.

Adds HEALTHCHECK to the `Dockerfile` runtime stage:
  --interval=60s --timeout=10s --start-period=30s --retries=3
  CMD plextraktsync healthcheck

Closes #2323

```
plextraktsync                 running   Up 36 seconds (health: starting)
[...]
plextraktsync                 running   Up 3 minutes (healthy)
```

### DIFF
--- a/plextraktsync/commands/healthcheck.py
+++ b/plextraktsync/commands/healthcheck.py
@@ -1,0 +1,53 @@
+from __future__ import annotations
+
+import re
+import sys
+import time
+from datetime import datetime
+
+from plextraktsync.factory import factory
+
+# Matches log level in the file format: "%(asctime)-15s %(levelname)s[%(name)s]:%(message)s"
+ERROR_RE = re.compile(r"\b(?:ERROR|CRITICAL)\[")
+LOG_TIMESTAMP_RE = re.compile(r"^(\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2})")
+
+_TAIL_BYTES_PER_SECOND = 256
+
+
+def healthcheck(max_age: int):
+    log_file = factory.config.log_file
+    cutoff = time.time() - max_age
+    last_error = None
+
+    tail_bytes = max_age * _TAIL_BYTES_PER_SECOND
+
+    try:
+        with open(log_file, "rb") as f:
+            f.seek(0, 2)
+            size = f.tell()
+            if size > tail_bytes:
+                f.seek(size - tail_bytes)
+                f.readline()  # discard partial line at seek point
+            content = f.read().decode("utf-8", errors="replace")
+    except FileNotFoundError:
+        print(f"Log file not found: {log_file}", file=sys.stderr)
+        raise SystemExit(1)
+
+    for line in content.splitlines():
+        if not ERROR_RE.search(line):
+            continue
+        m = LOG_TIMESTAMP_RE.match(line)
+        if m:
+            try:
+                ts = datetime.strptime(m.group(1), "%Y-%m-%d %H:%M:%S")
+                if ts.timestamp() < cutoff:
+                    continue
+            except ValueError:
+                pass  # unparseable timestamp: treat as recent
+        last_error = line.strip()
+
+    if last_error:
+        print(f"Healthcheck failed: {last_error}", file=sys.stderr)
+        raise SystemExit(1)
+
+    print("Healthcheck passed")
--- a/README.md
+++ b/README.md
@@ -136,6 +136,17 @@ services:
       - TZ=Europe/Tallinn
 ```
 
+To enable optional container health monitoring, add a `healthcheck` entry. This scans the log for ERROR or CRITICAL entries within the last hour and marks the container unhealthy if any are found.
+
+```yaml
+    healthcheck:
+      test: ["CMD", "plextraktsync", "healthcheck"]
+      interval: 60s
+      timeout: 10s
+      start_period: 30s
+      retries: 3
+```
+
 You can use specific version `0.25.16`:
 
 - `image: ghcr.io/taxel/plextraktsync:0.25.16`
--- a/plextraktsync/cli.py
+++ b/plextraktsync/cli.py
@@ -331,6 +331,17 @@ def self_update():
 
 
 @command()
+@click.option("--max-age", type=int, default=3600, show_default=True, help="Only consider log entries newer than N seconds")
+def healthcheck():
+    """
+    Check health of PlexTraktSync services.
+
+    Scans the log for ERROR or CRITICAL entries within --max-age seconds.
+    Exits 0 if healthy, 1 if any errors are found.
+    """
+
+
+@command()
 def bug_report():
     """
     Create a pre-populated GitHub issue with information about your configuration
@@ -369,6 +380,7 @@ def config():
 
 
 cli.add_command(bug_report)
+cli.add_command(healthcheck)
 cli.add_command(cache)
 cli.add_command(clear_collections)
 cli.add_command(compare_libraries)


glensc:
# CodeReview

Lint fails. Setup yourself git hooks:
- https://github.com/Taxel/PlexTraktSync/blob/1cd7806e54d5295bc93defaf0e81b3c74e10b045/CONTRIBUTING.md#git-setup-pre-commit

Also, I prefer atomic commits:
- https://cbea.ms/git-commit/

Can you split your commits?

glensc:
# Usefulness

Monitoring log for update doesn't mean the service is healthy... It might be stuck, doing nothing

DennisGaida:
> Usefulness

Fair point - this is a health check, not a liveness check. It won't catch a stuck/idle process. The primary motivation was issue #2323: watch logs ERROR/CRITICAL entries when auth fails (e.g. OAuth token expired) but the container stays Up with no way for Docker to know something is wrong. This check catches that specific failure mode.

> Lint fails & Can you split your commits?

Fixed Lint and split commits as requested.

plextraktsync/commands/healthcheck.py
```diff
@@ -0,0 +1,55 @@
...
+                        continue
+                except ValueError:
+                    pass  # unparseable timestamp: treat as recent
+            last_error = line.strip()
+
+    if last_error:
+        logger.error(f"Healthcheck failed: {last_error}")
+        sys.exit(1)
+
+    logger.info("Healthcheck passed")
```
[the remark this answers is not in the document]
DennisGaida: Fixed - switched all output to print()

plextraktsync/commands/healthcheck.py
```diff
@@ -0,0 +1,55 @@
...
+        logger.error(f"Log file not found: {log_file}")
+        sys.exit(1)
+
+    with f:
+        tail_bytes = max_age * _TAIL_BYTES_PER_SECOND
+        f.seek(0, 2)
+        size = f.tell()
+        if size > tail_bytes:
+            f.seek(size - tail_bytes)
+            f.readline()  # discard partial line at seek point
```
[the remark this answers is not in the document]
DennisGaida: Fixed - the file is now opened in binary mode ("rb")

Dockerfile
```diff
@@ -112,6 +112,9 @@ RUN ln -s /app/plextraktsync.sh /usr/bin/plextraktsync
 # https://github.com/python/cpython/issues/69667
 RUN chmod a+x /root
 
+HEALTHCHECK --interval=60s --timeout=10s --start-period=30s --retries=3 \
```
glensc: Let's have this optional for now.

Add as documentation to `README.md `how to update `docker-compose.yml`

plextraktsync/commands/healthcheck.py
```diff
@@ -0,0 +1,53 @@
...
+        with open(log_file, "rb") as f:
+            f.seek(0, 2)
+            size = f.tell()
+            if size > tail_bytes:
+                f.seek(size - tail_bytes)
+                f.readline()  # discard partial line at seek point
+            content = f.read().decode("utf-8", errors="replace")
+    except FileNotFoundError:
+        print(f"Log file not found: {log_file}", file=sys.stderr)
+        sys.exit(1)
```
glensc:
```suggestion
        raise SystemExit("Boom")
```

let's not call `sys.exit` directly

glensc:
e3d2569b13b83404897647ec8603f8bbaa5e3e50 could be split further :) 
1. adding command
2. integrating command

Dockerfile
DennisGaida: Added to README in 9b01adc17bebf1dabd35e0440a8f55526b072db5

plextraktsync/commands/healthcheck.py
DennisGaida: [        raise SystemExit(1)](https://github.com/Taxel/PlexTraktSync/pull/2478/changes/b1cbea5a1addedad311ee38037f289da4667a8cb#diff-1d32aaedb7b572e95380fc885ce2d7f52d73c45cfc935fc5b1a6b954160ed8d9R51)

DennisGaida:
> [e3d2569](https://github.com/Taxel/PlexTraktSync/commit/e3d2569b13b83404897647ec8603f8bbaa5e3e50) could be split further :)

When you say atomic, you **really** mean atomic 😉 

split into 3 commits, removed the hardcoded healthcheck in the dockerfile and added some doc to readme

simonc56:
Is logs scanning a standard way to do healthcheck and terminate a program ?

Shouldn't the python code just exit with code 1 when critical errors appear ?

DennisGaida:
> Is logs scanning a standard way to do healthcheck and terminate a program ?

There's no single standard (see https://docs.docker.com/reference/dockerfile#healthcheck). A Docker `healthcheck` is whatever you define as healthy - the most common pattern for HTTP services is simply `curl /ping`, which tells you nothing about whether the service is actually working, only that the process is alive and listening. Log scanning is a reasonable equivalent for a non-HTTP daemon: it signals that the process is running without errors, which is the same bar most `/ping` endpoints set.  
The check doesn't imply that a restart will fix the problem, it simply surfaces the unhealthy state to your orchestrator so it becomes visible and actionable (alert, restart policy, etc.) rather than silently failing (which we have happening right now). Multiple errors in the log over some amount of time means "unhealthy" to me personally.

> Shouldn't the python code just exit with code 1 when critical errors appear ?

Ideally yes, and that would be a better long-term fix for [#2323](https://github.com/Taxel/PlexTraktSync/issues/2323). But that requires finding every silent failure path in `watch` (as well as `sync`) and making it propagate errors correctly, a more invasive change. This PR adds observable health state today, without touching the core watch logic. Both approaches are complementary, not mutually exclusive.

Back to my initial point: I want my containers to have a sensible `healthcheck` to glance at them in some monitoring tool and see unhealthy containers and do something with them. This PR adds a very simple healthcheck based on the logic multiple errors=unhealthy.

glensc:
> > [e3d2569](https://github.com/Taxel/PlexTraktSync/commit/e3d2569b13b83404897647ec8603f8bbaa5e3e50) could be split further :)
> 
> When you say atomic, you **really** mean atomic 😉
> 
> split into 3 commits, removed the hardcoded healthcheck in the dockerfile and added some doc to readme

I know, I also struggle defining the logic to LLM agents, any suggestions?  :) 

Why: even the change would not be fully atomic, I find it easier to work with such history especially whey trying to split the pull request changes to different pull requests.

glensc:
I'm fine with the change as it's non-intrusive for all users.

The check could also be implemented in a way that the watch command listens for http port (enabled only with a flag), this could be done in a separate PR.

And if proper error propagation is added to the code, maybe not even need healthcheck at all.

glensc:
Also, generic contribution suggestion:

Avoid creating PR from your main branch:

> @DennisGaida wants to merge 3 commits into [Taxel:main](https://github.com/Taxel/PlexTraktSync/tree/main) from [DennisGaida:main](https://github.com/DennisGaida/PlexTraktSync/tree/main)

- https://github.com/Taxel/PlexTraktSync/blob/8d3f0aff61089faa9081342d334052b35cd6150b/CONTRIBUTING.md#we-use-github-flow-so-all-code-changes-happen-through-pull-requests

You maybe don’t acknowledge the problem, but all your **main** branch changes end up in an active pull request, so unintentional changes may appear there.

DennisGaida:
Nice!

> I know, I also struggle defining the logic to LLM agents, any suggestions? :)

Not a good rule, but I guess telling your LLM agents to be as atomic as possible is a good rule of thumb. You'd only have to take care that it doesn't create a commit per line-change ;-)

> Avoid creating PR from your main branch

Yeah, usually I don't. I'm actually going to delete my fork so it doesn't really matter what branch. It would be prettier for sure to have created a feature branch.

PR #2500, opened 2026-06-12
fix: username_filter fails closed on transient sessions probe failure

### LINKED ISSUE #2499: watch: username_filter fails open permanently if sessions probe fails once at startup
### Confirmation

- [x] I have tried downgrading to find version that can be used as a workaround

### The problem

With `watch.username_filter: true` (the default), the username filter is silently and **permanently disabled** for the lifetime of the watch process if a single `server.sessions()` probe fails when the filter is first evaluated. From that point on, every user's playback on the server is scrobbled to the owner's Trakt account. The only trace is a single WARNING-level log line.

In my case this ran undetected for 4 days and scrobbled 64 history entries from ~10 Plex users into my Trakt account.

Root cause is in `plextraktsync/watch/WatchStateUpdater.py`:

```python
@cached_property
def username_filter(self):
    if not self.config["watch"]["username_filter"]:
        return None

    if self.plex.has_sessions():
        # This must be a username, not email
        return self.plex.account.username

    self.logger.warning("No permission to access sessions, disabling username filter")
    return None
```

combined with `plextraktsync/plex/PlexApi.py`:

```python
def has_sessions(self):
    try:
        self.server.sessions()
        return True
    except (Unauthorized, BadRequest):
        return False
```

Three compounding problems:

1. **Fail-open**: a probe failure disables *filtering* instead of disabling *scrobbling*. `can_scrobble()` returns `True` for all sessions when `username_filter` is `None`, so the failure state is the opposite of the safe state.
2. **Cached**: `@cached_property` means one transient `Unauthorized`/`BadRequest` — evaluated lazily on the first playback event, i.e. exactly when PMS may still be starting up — locks the filter off until the container is recreated. A transient error is treated as a permanent "no permission" condition.
3. **Invisible**: the message is WARNING-level, so `plextraktsync healthcheck` (which matches only `ERROR|CRITICAL`) reports healthy while scrobbling all users unfiltered.

Beyond polluting the owner's history, this is a privacy issue in both directions: other household users' viewing is exposed to the owner's public Trakt profile without anyone's knowledge.

1. Docker deployment per README, `command: watch`, admin/owner Plex token, `watch.username_filter: true`, multiple users on the PMS.
2. Ran fine for a long period (filter working, only my plays scrobbled).
3. On 2026-06-08 an automated container update tool updated the Plex Media Server container; PMS was down/starting for ~2 minutes.
4. The PTS websocket connection died and the PTS container was restarted by Docker while PMS was still initialising (PMS can return 401/400 on `/status/sessions` during startup).
5. On the first playback event after restart, `has_sessions()` failed once → warning logged → filter permanently disabled.
6. All subsequent plays from all ~10 users scrobbled to my Trakt account for 4 days, until I noticed foreign history.

Reproduce deterministically: start `plextraktsync watch` while PMS is unreachable or still booting (so the first sessions probe raises `Unauthorized`/`BadRequest`), then play media as any non-owner user once PMS is up. Their play is scrobbled to the owner's Trakt account, and continues to be even after PMS is fully healthy.

Cleanup required manual removal of 64 history entries via the Trakt API (`/sync/history/remove`) followed by a full `sync` to restore my own watched state.

### Error trace / logs

```text
WARNING  No permission to access sessions, disabling username filter
```

### Expected behavior

A configured username filter should never silently degrade into "scrobble everyone". Any of the following would be acceptable, roughly in order of preference:

1. **Don't cache failure** — cache the filter only on a successful probe; on probe failure, skip scrobbling for that event and retry on the next one. A transient PMS startup error then costs nothing.
2. **Fail closed** — if the filter is configured but cannot be established, refuse to scrobble (or exit non-zero so Docker restart policies and healthchecks engage) rather than scrobbling unfiltered.
3. At minimum, log at ERROR (so `plextraktsync healthcheck` flags it) and re-log periodically while running unfiltered, instead of a single WARNING at the moment of failure.

```text
Not item-specific — no media matching problem. The affected entries were ordinary, correctly-matched items belonging to other users' sessions.
```

### Workarounds

No version workaround — the fail-open `cached_property` logic is present in current `main`, so downgrading doesn't help.

Operational mitigations I'm using/considered:
- Recreate the container after any PMS restart/update so the filter re-arms (manual, easy to forget).
- A custom Docker healthcheck that greps the log for "disabling username filter" newer than process start, paired with autoheal — works, but is fail-open for the detection interval and breaks silently if the log message wording changes.
- Patching `username_filter` to raise instead of returning None (fail-closed) — works, but has to be re-applied across image updates.

None of these are good substitutes for an upstream fix.

### Config file contents

```yaml
# Config File: /app/config/config.yml
cache:
  path: /app/config/trakt_cache
excluded-libraries:
- Private
- Family Holidays
config:
  dotenv_override: true
plex:
  timeout: 30
logging:
  append: true
  console_time: false
  debug: false
  filename: plextraktsync.log
  rotation:
    enable: true
    max_bytes: 10485760
    backup_count: 5
  filter_loggers: null
  filter: null
sync:
  rating_priority: plex
  plex_to_trakt:
    collection: false
    clear_collected: false
    ratings: true
    watched_status: true
    watchlist: true
  trakt_to_plex:
    liked_lists: true
    ratings: true
    watched_status: true
    watchlist: true
    watchlist_as_playlist: false
    playback_status: false
liked_lists:
  keep_watched: true
watch:
  add_collection: false
  remove_collection: false
  username_filter: true
  media_progressbar: true
  ignore_clients: null
xbmc-providers:
  movies: imdb
  shows: tvdb
```

### Install method

pipx

### Version

0.35.7

### Python Version

3.13.10

### Plex Server Version

1.43.2.10687-563d026ea

Ubuntu 24.10 - Docker

rageebm:
Fixes #2499 

Changes `username_filter` from `@cached_property` to `@property` with manual
success caching so a transient `has_sessions()` failure during PMS startup
doesn't permanently disable user filtering for the process lifetime.

**What changed:**

- `username_filter` retries the sessions probe on each event instead of
  caching the first failure
- Owner sessions (no `PLEX_ACCOUNT_TOKEN`): fail closed — skip scrobbling
  until sessions access recovers, log at ERROR once, log recovery at INFO
- Non-owner sessions (`PLEX_ACCOUNT_TOKEN` set): preserve existing behavior
  (warn once, run unfiltered — safe because their event stream only carries
  their own sessions)
- `sessions` and `can_scrobble` guard on the static config flag
  (`_username_filter_enabled`) instead of the dynamic `username_filter` value,
  preventing a secondary caching bug where `SessionCollection` caches as
  `None` during a transient outage

**Backward compatibility:** zero behavior change for any working setup.
The only path that behaves differently is owner token + transient probe
failure, which previously scrobbled all users unfiltered.

### DIFF
--- a/plextraktsync/watch/WatchStateUpdater.py
+++ b/plextraktsync/watch/WatchStateUpdater.py
@@ -39,6 +39,9 @@ class WatchStateUpdater(SetWindowTitle):
         self.remove_collection = config["watch"]["remove_collection"]
         self.add_collection = config["watch"]["add_collection"]
         self.session_media = {}
+        self.username_filter_value: str | None = None
+        self.username_filter_resolved = False
+        self.sessions_probe_warned = False
 
     def clamp_percent(self, percent: float) -> float:
         if percent < 0:
@@ -48,16 +51,42 @@ class WatchStateUpdater(SetWindowTitle):
         return percent
 
     @cached_property
-    def username_filter(self):
-        if not self.config["watch"]["username_filter"]:
+    def username_filter_enabled(self) -> bool:
+        return bool(self.config["watch"]["username_filter"])
+
+    @cached_property
+    def is_owner_session(self) -> bool:
+        """Non-owner (shared-server) users authenticate via PLEX_ACCOUNT_TOKEN."""
+        return not self.config["PLEX_ACCOUNT_TOKEN"]
+
+    @property
+    def username_filter(self) -> str | None:
+        if not self.username_filter_enabled:
             return None
+
+        if self.username_filter_resolved:
+            return self.username_filter_value
 
         if self.plex.has_sessions():
             # This must be a username, not email
-            return self.plex.account.username
+            self.username_filter_value = self.plex.account.username
+            self.username_filter_resolved = True
+            if self.sessions_probe_warned:
+                self.logger.info("Sessions access recovered, username filter active")
+        elif self.is_owner_session:
+            # Owner token: transient failure during PMS startup — do not cache,
+            # retry on the next event and skip scrobbling meanwhile (fail closed).
+            # See https://github.com/Taxel/PlexTraktSync/issues/2499
+            if not self.sessions_probe_warned:
+                self.logger.error("No permission to access sessions, will retry (skipping events until accessible)")
+                self.sessions_probe_warned = True
+        else:
+            # Non-owner tokens can't access sessions by design; run unfiltered
+            # (safe because their event stream only carries their own sessions)
+            self.logger.warning("No permission to access sessions, disabling username filter")
+            self.username_filter_resolved = True
 
-        self.logger.warning("No permission to access sessions, disabling username filter")
-        return None
+        return self.username_filter_value
 
     @cached_property
     def ignore_clients(self):
@@ -74,7 +103,7 @@ class WatchStateUpdater(SetWindowTitle):
 
     @cached_property
     def sessions(self):
-        if not self.username_filter:
+        if not self.username_filter_enabled:
             return None
 
         from plextraktsync.plex.SessionCollection import SessionCollection
@@ -221,10 +250,16 @@ class WatchStateUpdater(SetWindowTitle):
             if event.client_identifier in self.ignore_clients:
                 return False
 
-        if not self.username_filter:
+        if not self.username_filter_enabled:
             return True
 
-        return self.sessions[event.session_key] == self.username_filter
+        username = self.username_filter
+        if username is None:
+            # Non-owner: _username_filter_resolved is True → scrobble (safe)
+            # Owner: _username_filter_resolved is False → skip (fail closed)
+            return self.username_filter_resolved
+
+        return self.sessions[event.session_key] == username
 
     def scrobble(self, m: Media, percent: float, event: PlaySessionStateNotification):
         tm = m.trakt


plextraktsync/watch/WatchStateUpdater.py
```diff
@@ -48,16 +51,42 @@ def clamp_percent(self, percent: float) -> float:
...
             # This must be a username, not email
-            return self.plex.account.username
+            self._username_filter = self.plex.account.username
+            self._username_filter_resolved = True
+            if self._sessions_probe_warned:
+                self.logger.info("Sessions access recovered, username filter active")
+        elif self._is_owner_session:
+            # Owner token: transient failure during PMS startup — do not cache,
+            # retry on the next event and skip scrobbling meanwhile (fail closed).
+            # See https://github.com/Taxel/PlexTraktSync/issues/XXXX
```
glensc: Fix or remove the placeholder

plextraktsync/watch/WatchStateUpdater.py
```diff
@@ -39,6 +39,9 @@ def __init__(
         self.remove_collection = config["watch"]["remove_collection"]
         self.add_collection = config["watch"]["add_collection"]
         self.session_media = {}
+        self._username_filter: str | None = None
+        self._username_filter_resolved = False
+        self._sessions_probe_warned = False
```
glensc: remove the underscroes from property names please

glensc (changes requested):

glensc:
if this is ai-generated, express it clearly in pr body

rageebm:
> if this is ai-generated, express it clearly in pr body

It is - it's a fix I did for myself using claude.ai. I disclosed this in the bug report but not PR, my mistake.

Other requested changes have been implemented, thank you.

plextraktsync/watch/WatchStateUpdater.py
rageebm: Done

plextraktsync/watch/WatchStateUpdater.py
rageebm: Done

glensc:
approved ci run,

Meanwhile, @rageebm, please resolve discussions

image

glensc:
Released as 0.35.11

PR #2512, opened 2026-06-30
feat: add newer-wins rating priority

rcortese:
## Summary
- Adds `rating_priority: newer_wins` for rating sync.
- Uses rating timestamps to decide which side wins when Plex and Trakt ratings differ.
- Keeps the existing `plex`, `trakt`, and `none` behavior unchanged.
- Documents the new option in `config.default.yml`.

## Behavior
With `rating_priority: newer_wins`:
- if both sides have different ratings, the side with the newest `rated_at` / `lastRatedAt` wins;
- if only Plex has a rating, it can fill Trakt when `plex_to_trakt.ratings` is enabled;
- if only Trakt has a rating, it can fill Plex when `trakt_to_plex.ratings` is enabled;
- if ratings are equal, no action is taken;
- dry-run still logs without writing.

If both timestamps are missing, Plex is treated as the newer side to preserve the previous default preference when both directions are enabled.

## Test Plan
- `python -m ruff check plextraktsync/sync/SyncRatingsPlugin.py tests/test_sync_ratings_plugin.py`
- YAML parse of `plextraktsync/config.default.yml`
- `python -m pytest -q tests/test_rating.py tests/test_sync_ratings_plugin.py` → `10 passed`
- `python -m pytest -q` → `32 passed, 11 skipped`

Note: the full test run prints an expected fake-API-key error log from existing tests, but exits successfully.

### DIFF
--- a/plextraktsync/config.default.yml
+++ b/plextraktsync/config.default.yml
@@ -54,10 +54,11 @@ logging:
 # settings for 'sync' command
 sync:
   # Setting for whether ratings from one platform should have priority.
-  # Valid values are trakt, plex or none. (default: plex)
+  # Valid values are trakt, plex, newer_wins or none. (default: plex)
   # none - No rating priority. Existing ratings are not overwritten.
   # trakt - Trakt ratings have priority. Existing Plex ratings are overwritten.
   # plex - Plex ratings have priority. Existing Trakt ratings are overwritten.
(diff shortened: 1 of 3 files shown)


rcortese:
Closing for now: this needs broader operational validation before being proposed upstream.

PR #2517, opened 2026-07-04
fix: handle None seasons/episodes/shows in pytrakt_extensions

### LINKED ISSUE #2515: Crash in 0.35.11: TypeError "'NoneType' object is not iterable" in watched-shows / sync
### Confirmation

- [ ] I have tried downgrading to find version that can be used as a workaround

### Version

PlexTraktSync 0.35.11
Docker

### Problem

Both commands fail:

```
plextraktsync watched-shows
plextraktsync sync
```

with:

```
TypeError: 'NoneType' object is not iterable
```

The traceback points to:

```python
self.seasons = {}
allCompleted = True

for season in seasons:
```

inside `pytrakt_extensions.py`.

### Investigation

I verified the following:

- Trakt authentication is valid.
- Plex connection is valid.
- `healthcheck` passes.
- `unmatched` works.
- Only `watched-shows` and `sync` crash.

I queried Trakt directly from inside the container:

```
GET /sync/watched/shows
```

The request returns HTTP 200.

The returned objects do not contain a `seasons` field, so `ShowProgress.__init__()` receives:

```python
seasons=None
```

which causes:

```python
for season in seasons:
```

to crash.

### Temporary workaround

Adding

```python
seasons = seasons or []
```

before the loop fixes the crash:

```python
self.seasons = {}
seasons = seasons or []
allCompleted = True

for season in seasons:
```

After this change:

- `watched-shows` works
- `sync` completes successfully

### Remaining issue

Every show is reported with:

```
Seasons: 0
```

so the workaround only prevents the crash.

It looks like the code no longer receives the season data it expects and probably needs additional handling.

trakt sync

### Error trace / logs

```text

```

### Expected behavior

seasons synced

```text

```

### Workarounds

_No response_

### Config file contents

```yaml

```

### Install method

docker-compose

### Version

latest as of july 3rd 2027

### Python Version

?

### Plex Server Version

latest QNAP qpkg

QTS latest

cjtylr:
Trakt's /sync/watched/shows (and /sync/collection/shows) endpoints can return an entry where the "seasons" key is null instead of an empty list (and, more rarely, episodes or the top-level shows list can be null too). This caused:

    TypeError: 'NoneType' object is not iterable

in ShowProgress.__init__, SeasonProgress.__init__, and AllShowsProgress.__init__, crashing the whole sync as soon as such a show was encountered.

Guard each with "x = x or []" before iterating, matching the pattern already used for optional dict fields like "show" in ShowProgress.

Fixes #2515

### DIFF
--- a/plextraktsync/pytrakt_extensions.py
+++ b/plextraktsync/pytrakt_extensions.py
@@ -46,6 +46,7 @@ class SeasonProgress:
         self.number = number
         self.aired = aired
         self.episodes = {}
+        episodes = episodes or []
         for episode in episodes:
             prog = EpisodeProgress(**episode)
             self.episodes[prog.number] = prog
@@ -93,6 +94,7 @@ class ShowProgress:
         self.slug = show["ids"]["slug"] if show else None
         self.seasons = {}
         allCompleted = True
+        seasons = seasons or []
         for season in seasons:
             prog = SeasonProgress(**season)
             self.seasons[prog.number] = prog
@@ -112,6 +114,7 @@ class ShowProgress:
 class AllShowsProgress:
     def __init__(self, shows=None):
         self.shows = {}
+        shows = shows or []
         for show in shows:
             prog = ShowProgress(**show)
             self.shows[prog.trakt] = prog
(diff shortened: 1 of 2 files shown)


cjtylr:
This fix not needed; needs to be fixed elsewhere to adapt to new Trakt API structure. See: https://github.com/Taxel/PlexTraktSync/issues/2515#issuecomment-4881465363

PR #2518, opened 2026-07-04
Fix: Paginate watched/collection with extended=progress

gaborbernat:
Trakt paginated its `sync/watched/{type}` and `sync/collection/{type}` endpoints during 2026 and stopped returning the per-season episode breakdown unless the caller asks for `extended=progress` (trakt/trakt-api#775). `allwatched()` still requested `sync/watched/shows` with no parameters, so every show arrived with `seasons: null`, `ShowProgress` raised `'NoneType' object is not iterable`, and `sync` aborted for anyone with watched shows. That is the crash reported in #2515. 🐛

The fix requests each page with `extended=progress` plus `page` and `limit`, then walks pages until Trakt returns a short one, because `extended=progress` caps the page size at 100. `allcollected` takes the same shape. The progress parsers now treat a missing `seasons`, `episodes`, or `shows` array as empty instead of raising, so a later change to the response shape degrades rather than crashing. ✨

I verified this against an account with 189 watched shows across two pages: stock `0.35.11` crashes on the first show, and with the change `sync` finishes with no traceback and reconciles per-episode watched state in both directions. `allcollected` uses `extended=metadata`, which I could not confirm against a populated Trakt collection since the discussion does not document the breakdown parameter for that endpoint; the null guards keep it from crashing regardless, and I can switch the value if a maintainer knows the right one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plextraktsync/pytrakt_extensions.py
+++ b/plextraktsync/pytrakt_extensions.py
@@ -1,21 +1,15 @@
 from __future__ import annotations
 
-from trakt.core import get
+from trakt.pagination import paginate
 from trakt.utils import airs_date
 
 
-@get
 def allwatched():
-    # returns an AllShowProgress object containing all watched shows
-    data = yield "sync/watched/shows"
-    yield AllShowsProgress(data)
+    return AllShowsProgress(paginate("sync/watched/shows", extended="progress"))
 
 
-@get
 def allcollected():
-    # returns an AllShowProgress object containing all collected shows
-    data = yield "sync/collection/shows"
-    yield AllShowsProgress(data)
+    return AllShowsProgress(paginate("sync/collection/shows", extended="metadata"))
 
 
 class EpisodeProgress:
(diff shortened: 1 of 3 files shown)


XenorPLxx:
Looks like it's similar case for lists.
I've verified the patch and it synces watched statuses correctly, but pulling trakt lists into plex playlists is limited to 100 items.

glensc:
Updated to require pytrakt:
- https://github.com/Taxel/PlexTraktSync/pull/2522

And paginate from 4.4.1:
- https://github.com/Taxel/PlexTraktSync/pull/2524

glensc:
The empty value guard probably was not needed, it seems it was carried from this pr:
- https://github.com/Taxel/PlexTraktSync/pull/2517/changes

But as this pr has no mention of this anywhere (pr, commit), I had no idea.

